PR #26, opened 2019-11-01
feat: preparing to support Directx12

sindharta:
Use D3D11.1 resource as an intermediate between DirectX12 and NVEncoder


sindharta:
Wrong target branch.

PR #27, opened 2019-11-01
chore: preparing to support Directx12

sindharta:
Use D3D11.1 as an intermediate between DX12 and NVEncode


Plugin~/WebRTCPlugin/GraphicsDevice/GraphicsDevice.cpp
```diff
@@ -30,9 +30,11 @@ void GraphicsDevice::Init(IUnityInterfaces* unityInterface) {
         }
         default: {
             DebugError("Unsupported Unity Renderer: %d", m_rendererType);
+            return;
             break;
```
karasusan: It seems better to remove `break` statement.

karasusan:
👍

PR #35, opened 2019-11-21
refactor removing duplicate ITexture2D creation calls when allocating input resource

sindharta:


Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoderGL.h
```diff
@@ -7,7 +7,6 @@ namespace WebRTC {
         NvEncoderGL(uint32_t nWidth, uint32_t nHeight, IGraphicsDevice* device);
         virtual ~NvEncoderGL();
     protected:
-        virtual void* AllocateInputBuffer() override;
-        virtual ITexture2D* CreateTexture2DFromInputBuffer(void* buffer) override;
+        void* AllocateInputResourceV(ITexture2D* tex);
```
karasusan: Is this override method like [that](https://github.com/Unity-Technologies/com.unity.webrtc/pull/35/files#diff-9fb3fda3bdf05a01cde4df19d4a78e94R11)?

sindharta: Yes, this is a virtual method. I'll add override keyword

PR #38, opened 2019-11-29
Feat/swencoder

koseyile:
Feature: software encoder.

### DIFF
--- a/Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
+++ b/Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
@@ -8,8 +8,8 @@
 
 #if defined(SUPPORT_D3D11)
 #include "NvCodec/NvEncoderD3D11.h"
-#include "SoftwareCodec/SoftwareEncoder.h"
 #endif
+#include "SoftwareCodec/SoftwareEncoder.h"
 
 #include "GraphicsDevice/IGraphicsDevice.h"
 
--- a/Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
+++ b/Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
@@ -1,5 +1,6 @@
 ﻿#include "pch.h"
 #include "IEncoder.h"
+#include "Context.h"
 #include "EncoderFactory.h"
 
 #if defined(SUPPORT_OPENGL_CORE)
@@ -31,11 +32,12 @@ namespace WebRTC {
         switch (deviceType) {
 #if defined(SUPPORT_D3D11)
             case GRAPHICS_DEVICE_D3D11: {
-#if USE_SOFTWARE_ENCODER
-                m_encoder = std::make_unique<SoftwareEncoder>(width, height, device);
-#else
-                m_encoder = std::make_unique<NvEncoderD3D11>(width, height, device);
-#endif // USE_SOFTWARE_ENCODER
+                if (!ContextManager::s_use_software_encoder)
+                {
+                    m_encoder = std::make_unique<NvEncoderD3D11>(width, height, device);
+                } else {
+                    m_encoder = std::make_unique<SoftwareEncoder>(width, height, device);
+                }
                 break;
             }
 #endif
--- a/Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
+++ b/Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
@@ -8,6 +8,7 @@
 
 #if defined(SUPPORT_D3D11)
 #include "NvCodec/NvEncoderD3D11.h"
+#include "SoftwareCodec/SoftwareEncoder.h"
 #endif
 
 #include "GraphicsDevice/IGraphicsDevice.h"
@@ -30,7 +31,11 @@ namespace WebRTC {
         switch (deviceType) {
 #if defined(SUPPORT_D3D11)
             case GRAPHICS_DEVICE_D3D11: {
+#if USE_SOFTWARE_ENCODER
+                m_encoder = std::make_unique<SoftwareEncoder>(width, height, device);
+#else
                 m_encoder = std::make_unique<NvEncoderD3D11>(width, height, device);
+#endif // USE_SOFTWARE_ENCODER
                 break;
             }
 #endif
--- a/Plugin~/WebRTCPlugin/Codec/SoftwareCodec/SoftwareEncoder.h
+++ b/Plugin~/WebRTCPlugin/Codec/SoftwareCodec/SoftwareEncoder.h
@@ -24,7 +24,7 @@ namespace WebRTC
     private:
         IGraphicsDevice* m_device;
         ITexture2D* m_encodeTex;
-        int width = 1920;
-        int height = 1080;
+        int m_width = 1920;
+        int m_height = 1080;
     };
 }
--- a/Plugin~/WebRTCPlugin/Codec/SoftwareCodec/SoftwareEncoder.h
+++ b/Plugin~/WebRTCPlugin/Codec/SoftwareCodec/SoftwareEncoder.h
@@ -1,0 +1,30 @@
+﻿#pragma once
+#include <vector>
+#include <thread>
+#include <atomic>
+#include "Codec/IEncoder.h"
+
+namespace WebRTC
+{
+    class IGraphicsDevice;
+    class ITexture2D;
+    class SoftwareEncoder : public IEncoder
+    {
+    public:
+        SoftwareEncoder(int _width, int _height, IGraphicsDevice* device);
+        void InitV() override;
+        void SetRate(uint32_t rate) override {}
+        void UpdateSettings() override {}
+        bool CopyBuffer(void* frame) override;
+        void EncodeFrame() override;
+        bool IsSupported() const override { return true; }
+        void SetIdrFrame() override {}
+        uint64 GetCurrentFrameCount() override { return 0; }
+
+    private:
+        IGraphicsDevice* m_device;
+        ITexture2D* m_encodeTex;
+        int width = 1920;
+        int height = 1080;
+    };
+}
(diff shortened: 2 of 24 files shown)


Plugin~/WebRTCPlugin/GraphicsDevice/D3D11/D3D11GraphicsDevice.cpp
```diff
@@ -76,4 +94,53 @@ bool D3D11GraphicsDevice::CopyResourceFromNativeV(ITexture2D* dest, void* native
     return true;
 }
 
+rtc::scoped_refptr<webrtc::I420Buffer> D3D11GraphicsDevice::ConvertRGBToI420(ITexture2D* tex)
+{
+    D3D11_MAPPED_SUBRESOURCE resource;
+    HRESULT hr = m_d3d11Context->Map((ID3D11Resource*)tex->GetNativeTexturePtrV(), 0, D3D11_MAP_READ, 0, &resource);
```
karasusan: please check result

Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
```diff
@@ -8,6 +8,7 @@
 
 #if defined(SUPPORT_D3D11)
 #include "NvCodec/NvEncoderD3D11.h"
+#include "SoftwareCodec/SoftwareEncoder.h"
```
sindharta: SoftwareEncoder by itself only depends on the base class and does not depend on the graphics implementation class (D3D11, etc), so I think we can put it outside ifdef

Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
```diff
@@ -30,7 +31,11 @@ namespace WebRTC {
         switch (deviceType) {
 #if defined(SUPPORT_D3D11)
             case GRAPHICS_DEVICE_D3D11: {
+#if USE_SOFTWARE_ENCODER
```
sindharta: Please use a variable instead of ifdef.
It will be great if we can set the variable from the Unity side, but it can also be a constant, for now.

Plugin~/WebRTCPlugin/Codec/SoftwareCodec/SoftwareEncoder.h
```diff
@@ -0,0 +1,30 @@
...
+        bool CopyBuffer(void* frame) override;
+        void EncodeFrame() override;
+        bool IsSupported() const override { return true; }
+        void SetIdrFrame() override {}
+        uint64 GetCurrentFrameCount() override { return 0; }
+
+    private:
+        IGraphicsDevice* m_device;
+        ITexture2D* m_encodeTex;
+        int width = 1920;
```
sindharta: member variables should use m_ something.

Plugin~/WebRTCPlugin/Context.cpp
```diff
@@ -127,19 +127,35 @@ namespace WebRTC
         audioDevice = new rtc::RefCountedObject<DummyAudioDevice>();
         nvVideoCapturerUnique = std::make_unique<NvVideoCapturer>();
         nvVideoCapturer = nvVideoCapturerUnique.get();
+
+#if USE_SOFTWARE_ENCODER
```
sindharta: Please use a variable instead of ifdef.
Also, the only difference seems to be 
- webrtc::CreateBuiltinVideoEncoderFactory() -> SoftwareEncoder
- dummyVideoEncoderFactory  -> NvEncoder

So maybe we can only do if for these two ?

Plugin~/WebRTCPlugin/GraphicsDevice/D3D11/D3D11GraphicsDevice.h
```diff
@@ -14,10 +14,12 @@ class D3D11GraphicsDevice : public IGraphicsDevice{
     virtual void ShutdownV();
     inline virtual void* GetEncodeDevicePtrV();
     virtual ITexture2D* CreateDefaultTextureV(uint32_t w, uint32_t h);
+    virtual ITexture2D* CreateStagingTextureV(uint32_t w, uint32_t h);
```
sindharta: "Staging" is a Direct3D11 term. I think we should rename it to something that represents what it actually does, perhaps: `CreateCPUReadTextureV`?

Plugin~/WebRTCPlugin/GraphicsDevice/IGraphicsDevice.h
```diff
@@ -15,11 +15,13 @@ class IGraphicsDevice {
     virtual bool InitV() = 0;
     virtual void ShutdownV() = 0;
     virtual ITexture2D* CreateDefaultTextureV(uint32_t width, uint32_t height) = 0;
+    virtual ITexture2D* CreateStagingTextureV(uint32_t width, uint32_t height) { return NULL; }
```
sindharta: returning NULL here defeats the purpose of making an interface class.
Rather, please write `return null` in each derived class and put assert or log so that it is clear that the implementation for that graphics device is not yet implemented.

Plugin~/WebRTCPlugin/GraphicsDevice/IGraphicsDevice.h
```diff
@@ -15,11 +15,13 @@ class IGraphicsDevice {
...
     virtual ITexture2D* CreateDefaultTextureFromNativeV(uint32_t width, uint32_t height, void* nativeTexturePtr) = 0;
     virtual void* GetEncodeDevicePtrV() = 0;
     virtual bool CopyResourceV(ITexture2D* dest, ITexture2D* src) = 0;
     virtual bool CopyResourceFromNativeV(ITexture2D* dest, void* nativeTexturePtr) = 0;
     virtual GraphicsDeviceType GetDeviceType() = 0;
+    virtual rtc::scoped_refptr<webrtc::I420Buffer> ConvertRGBToI420(ITexture2D* tex) { return NULL; }
```
sindharta: returning NULL here defeats the purpose of making an interface class.
Rather, please write `return null` in each derived class and put assert or log so that it is clear that the implementation for that graphics device is not yet implemented.

Plugin~/WebRTCPlugin/NvVideoCapturer.cpp
```diff
@@ -31,14 +31,32 @@ namespace WebRTC
         }
     }
 
-    void NvVideoCapturer::CaptureFrame(std::vector<uint8>& data)
+//    void NvVideoCapturer::CaptureFrame(std::vector<uint8>& data)
```
sindharta: Do we need these comments ?

Plugin~/WebRTCPlugin/Codec/EncoderFactory.cpp
koseyile: got it. this variable adds as a static member variable of ContextManager is ok?

sindharta: Ok

karasusan:
like this?  (Default is putting a checkmark)
☑️ Hardware encoder support

Plugin~/WebRTCPlugin/GraphicsDevice/D3D11/D3D11GraphicsDevice.cpp
koseyile: I have added assert and check code.

koseyile:
> 
> 
> like this? (Default is putting a checkmark)
> 
> ☑️ Hardware encoder support

yes. that's good.

Plugin~/WebRTCPlugin/Context.cpp
```diff
@@ -127,19 +127,35 @@ namespace WebRTC
...
+                                    nullptr);
+        }else{
+            peerConnectionFactory = webrtc::CreatePeerConnectionFactory(
+                                    workerThread.get(),
+                                    workerThread.get(),
+                                    signalingThread.get(),
+                                    audioDevice,
+                                    webrtc::CreateAudioEncoderFactory<webrtc::AudioEncoderOpus>(),
+                                    webrtc::CreateAudioDecoderFactory<webrtc::AudioDecoderOpus>(),
+                                    webrtc::CreateBuiltinVideoEncoderFactory(),
```
sindharta: Could you please use less code to do this branching, because the only difference between software/hardware encoder is
- std::move(dummyVideoEncoderFactory)
- webrtc::CreateBuiltinVideoEncoderFactory(),

koseyile: ok.

Plugin~/unity/include/IUnityGraphics.h
```diff
@@ -34,6 +34,12 @@ typedef enum UnityGfxDeviceEventType
     kUnityGfxDeviceEventAfterReset     = 3,
 } UnityGfxDeviceEventType;
 
+typedef enum UnityEncoderType
```
karasusan: This class has moved from the other repo. It is better not to touch it.
https://github.com/Unity-Technologies/NativeRenderingPlugin/tree/master/PluginSource/source/Unity

It is better to put this in `pch.h`.

Plugin~/WebRTCPlugin/GraphicsDevice/D3D12/D3D12GraphicsDevice.cpp
```diff
@@ -127,6 +132,12 @@ ITexture2D* D3D12GraphicsDevice::CreateSharedD3D12Texture(uint32_t w, uint32_t h
     return new D3D12Texture2D(w,h,nativeTex, handle, sharedTex);
 }
 
+rtc::scoped_refptr<webrtc::I420Buffer> D3D12GraphicsDevice::ConvertRGBToI420(ITexture2D* tex)
+{
+    assert(false && "ConvertRGBToI420 need to implement on D3D12");
+    return NULL;
```
karasusan: You can use `nullptr` instead of NULL.

PR #46, opened 2019-12-20
feat: Stats API

sindharta:


Runtime/Scripts/RTCPeerConnection.cs
```diff
@@ -330,6 +341,18 @@ static void OnFailureSetSessionDesc(IntPtr ptr)
                 connection.m_opSetDesc.Done();
             }, null);
         }
+
+        [AOT.MonoPInvokeCallback(typeof(DelegateSetSDFailure))]
```
karasusan: This should be used value `DelegateCollectStats` not `DelegateSetSDFailure`

WebRTC~/Packages/com.unity.webrtc/Tests
```diff
@@ -0,0 +1 @@
+../../../Tests
```
karasusan: I am not sure this file is needed

sindharta: I think we need this in order to have the tests inside the test project.
There is actually a better way to do this though, and I will fix this soon

Runtime/Scripts/RTCPeerConnection.cs
sindharta: Ok

karasusan:
LGTM!

PR #48, opened 2020-01-08
refactor: select encoder type (software or hardware)

karasusan:
### Changed
- Added `WebRTC.Initialize()` method argument `encoderType`

## Discussion
- Should we add the method `GetAvailableEncoderTypes`?
- How do users select hardware when the users use multiple GPU?


Plugin~/WebRTCPlugin/Context.cpp
```diff
@@ -8,17 +8,16 @@
 namespace WebRTC
 {
     ContextManager ContextManager::s_instance;
-    bool ContextManager::s_use_software_encoder = false;
 
-    Context* ContextManager::GetContext(int uid)
+    Context* ContextManager::GetContext(int uid, UnityEncoderType encoderType)
```
sindharta: Adding encoderType as a parameter here seems a bit strange because this function is supposed to return an existing context. 

If we want to create a default Context when it doesn't exist, I think it's better to just hardcode it somewhere rather than using a parameter.

karasusan: Sure, I guess we should define a method to create context instance like below
`ContextManager::CreateContext(int uid, UnityEncoderType encoderType)`

I will change the behavior of `GetContext` so it returns null if `context` instance does not exist.

Plugin~/WebRTCPlugin/Context.cpp
```diff
@@ -177,6 +184,16 @@ namespace WebRTC
         nvVideoCapturer->FinalizeEncoder();
     }
 
+    bool Context::SetEncoderType(UnityEncoderType type)
```
sindharta: Do we want to support changing the encoder type ?
If so, then we might need to do the initialization process again.

karasusan: we now cannot support this feature.
To avoid misunderstanding, I will remove this method.

karasusan:
The cause of macos build failed is that `prepack` process can not upload the macos plugin.

PR #50, opened 2020-01-10
feat: Mac support using Software Encoder

sindharta:


Tests/Editor/PluginTest.cs
```diff
@@ -15,6 +15,9 @@ public class PluginTest {
     [Test]
     [UnityPlatform(RuntimePlatform.WindowsEditor, RuntimePlatform.OSXEditor)]
```
karasusan: `UnityPlatformAttribute` can select platforms as test targets.
Please remove `RuntimePlatform.OSXEditor`

sindharta: Fixed

karasusan:
Ship it!

PR #71, opened 2020-02-14
update: libwebrtc to M79

kannan-xiao4:


karasusan:
Is it necessary to remove it?
https://github.com/Unity-Technologies/com.unity.webrtc/pull/71/files#diff-e6831f902f0633267c6ebec22e0c090e

PR #75, opened 2020-02-21
fix: delay load dlls for vulkan

karasusan:
- added delay load dlls (`nvcuda.dll`, `vulkan-1.dll`)
- added native API `NativeMethods.GetHardwareEncoderSupport`
- changed attribute to exclude tests for hardware encoder when tests run on the poor pc


sindharta:
LGTM

PR #83, opened 2020-03-19
Fix Color bug using OpenGL Encoder

kannan-xiao4:
Fix this issue https://github.com/Unity-Technologies/UnityRenderStreaming/issues/239


sindharta:
No problem. But we should fix the CI errors first before mergin

karasusan:
LGTM

PR #86, opened 2020-03-26
refactor: Changed the way of pass native string 

karasusan:
## Changed
- Add utility method `ConvertString`
- Changed `PeerConnectionObject::GetConfiguration()` method return value
- Add IntPtr extension method to free native memory

### DIFF
--- a/Runtime/Scripts/IntPtrExtension.cs
+++ b/Runtime/Scripts/IntPtrExtension.cs
@@ -1,0 +1,20 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Unity.WebRTC
+{
+
+    public static class IntPtrExtension
+    {
+        public static string AsAnsiString(this IntPtr ptr)
+        {
+            if (ptr == IntPtr.Zero)
+            {
+                throw new ArgumentException("ptr is nullptr");
+            }
+            string str = Marshal.PtrToStringAnsi(ptr);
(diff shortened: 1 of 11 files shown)


Runtime/Scripts/IntPtrExtension.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class IntPtrExtension
+    {
+        public static string AsAnsiString(this IntPtr ptr)
+        {
+            if (ptr == IntPtr.Zero)
+            {
+                throw new ArgumentException("ptr is nullptr");
+            }
+            string str = Marshal.PtrToStringAnsi(ptr);
+            Marshal.FreeHGlobal(ptr);
```
sindharta, line 16: I am not sure if calling `Marshall.FreeHGlobal()` here is safe. 
This assumes that all strings coming from C++ were allocated using `Marshall.GlobalAlloc()` in the heap, and therefore I am not sure if this will be safe for strings which are allocated in stack or static string variables.

sindharta: Also, the counterpart of `CoTaskMemAlloc`  is `FreeCoTaskMem`.

karasusan: Indeed, this code is unsafe if the memory allocated in stack.
Do you have a good idea to identify the memory allocated in heap or stack?

PR #95, opened 2020-04-12
fix: bitrate control

karasusan:
In this PR fixed bitrate issue

- `DummyVideoEncoder` receives events from webrtc API to control bitrate
- Define `IVideoEncoderObserver` class to detect events on `DummyVideoEncoder`
- `Context` class implements `IVideoEncoderObserver` and process the callback
- Remove `bitrate` parameter from MediaStreamTrack API because the feature not supported yet

## TODO
First, the original design has problems that the hardware encoder has been placed out of `DummyVideoEncoder`. It should be refactored but will take some time, so I don't fix it in this version.

`UnityVideoEncoderFactory` that factory class to make video encoders but in this current version only can support h264. It should be supported others like VP8, Vp9, and AV1.

### DIFF
--- a/Plugin~/WebRTCPlugin/DummyVideoEncoder.cpp
+++ b/Plugin~/WebRTCPlugin/DummyVideoEncoder.cpp
@@ -40,6 +40,8 @@ namespace webrtc
     int32_t DummyVideoEncoder::Release()
     {
         this->callback = nullptr;
+        this->m_setKeyFrame.disconnect_all();
+        this->m_setRates.disconnect_all();
         return WEBRTC_VIDEO_CODEC_OK;
     }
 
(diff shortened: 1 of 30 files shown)


karasusan:
I found a crash bug...
I am fixing it.

Plugin~/WebRTCPlugin/DummyVideoEncoder.cpp
```diff
@@ -7,94 +7,111 @@ namespace unity
...
+        return WEBRTC_VIDEO_CODEC_OK;
+    }
+
+    int32_t DummyVideoEncoder::RegisterEncodeCompleteCallback(webrtc::EncodedImageCallback* callback)
+    {
+        this->callback = callback;
+        return WEBRTC_VIDEO_CODEC_OK;
+    }
+
+    int32_t DummyVideoEncoder::Release()
```
sindharta, line 40: Can we release these too ?
```
        this->m_setKeyFrame.connect(observer, &IVideoEncoderObserver::SetKeyFrame);
        this->m_setRates.connect(observer, &IVideoEncoderObserver::SetRates);
```

karasusan: OK, will fix it

karasusan: Added these lines into `Release` method.
```
this->m_setKeyFrame.disconnect_all();
this->m_setRates.disconnect_all();
```

PR #98, opened 2020-04-21
chore: stop playing if using not support graphics api

kannan-xiao4:
OpenGL api not support on Windows / Mac.
So, I stop playing on that time.

### DIFF
--- a/Runtime/Scripts/WebRTC.cs
+++ b/Runtime/Scripts/WebRTC.cs
@@ -4,6 +4,7 @@ using UnityEngine;
 using System.Threading;
 using System.Collections;
 using System.Collections.Generic;
+using UnityEngine.Rendering;
 
 namespace Unity.WebRTC
 {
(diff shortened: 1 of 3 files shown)


Runtime/Scripts/WebRTC.cs
```diff
@@ -230,6 +231,23 @@ public static class WebRTC
...
+            {
+                if (SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLCore ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLES2 ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLES3)
+                {
+                    Debug.LogError($"Not Support OpenGL API on {Application.platform}");
+#if UNITY_EDITOR
+                    UnityEditor.EditorApplication.isPlaying = false;
+#else
+                    Application.Quit ();
```
karasusan, line 245: It should be better to throw the exception to notice the error to developers.

Runtime/Scripts/WebRTC.cs
```diff
@@ -230,6 +231,23 @@ public static class WebRTC
...
+                if (SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLCore ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLES2 ||
+                    SystemInfo.graphicsDeviceType == GraphicsDeviceType.OpenGLES3)
+                {
+                    Debug.LogError($"Not Support OpenGL API on {Application.platform}");
+#if UNITY_EDITOR
+                    UnityEditor.EditorApplication.isPlaying = false;
+                    return;
+#else
+                    throw new NotSupportedException();
```
karasusan, line 246: Good, `NotSupportedException` it is correct to use in this situation.
This could be better if you add a comment for telling what happened.

https://docs.microsoft.com/en-us/dotnet/api/system.notsupportedexception.-ctor?view=netframework-4.8#System_NotSupportedException__ctor_System_String_

karasusan:
LGTM

PR #111, opened 2020-05-08
Fix crash on encoder initialization

kannan-xiao4:


Runtime/Scripts/MediaStreamTrack.cs
```diff
@@ -108,6 +108,19 @@ internal VideoStreamTrack(string label, UnityEngine.Texture source, UnityEngine.
...
+        /// The videotrack cannot be used if the encoder has not been initialized.
+        /// Do not use it until the initialization is complete.
+        /// </summary>
+        public bool IsInitialized
+        {
+            get
+            {
+                return WebRTC.Context.GetInitializationResult(self) == CodecInitializationResult.Success;
+            }
+        }
```
kannan-xiao4, line 122: https://github.com/Unity-Technologies/com.unity.webrtc/blob/4b2125547d6c23f66720e4e388a6a5ed7bd5cf3b/Runtime/Scripts/MediaStreamTrack.cs#L176
this method work another thread (not main thread)
so, this flag need check next flame

karasusan: Could you add test cases for the method ?

kannan-xiao4: I'll add that.
However, I don't think I can write a test because the maximum number of initialization failures depends on the GPU. Because there is no API to get the maximum number.

karasusan: You are right, we don't need the test case as you said.
But, at least we can test that the status works correctly
- `NotInitialized`: Not started the initialization of the encoder
- `Success`: the initialization of the encoder succeeds.

kannan-xiao4: I add assert about new method and property.

b03ddc2

Tests/Runtime/NativeAPITest.cs
```diff
@@ -253,12 +253,17 @@ public IEnumerator CallVideoEncoderMethods()
             var sender = NativeMethods.PeerConnectionAddTrack(peer, track, streamId);
 
             var callback = NativeMethods.GetRenderEventFunc(context);
+            var initializeResult = NativeMethods.GetInitializationResult(context, track);
+            Assert.IsTrue(initializeResult == CodecInitializationResult.NotInitialized);
```
karasusan, line 257: I guess `Assert.AreEqual` is better than  `Assert.IsTrue`.
```
public static void AreEqual(object expected, object actual);
```

Tests/Runtime/NativeAPITest.cs
```diff
@@ -253,12 +253,17 @@ public IEnumerator CallVideoEncoderMethods()
...
 
             // TODO::
             // note:: You must call `InitializeEncoder` method after `NativeMethods.ContextCaptureVideoStream`
             NativeMethods.ContextSetVideoEncoderParameter(context, track, width, height);
             VideoEncoderMethods.InitializeEncoder(callback, track);
             yield return new WaitForSeconds(1.0f);
+            initializeResult = NativeMethods.GetInitializationResult(context, track);
+            Assert.IsTrue(initializeResult == CodecInitializationResult.Success);
```
karasusan, line 265: I guess `Assert.AreEqual` is better than  `Assert.IsTrue`.
```
public static void AreEqual(object expected, object actual);
```

PR #118, opened 2020-05-14
feat: Add API RTCRtpSender.SetParameters

karasusan:
This API provides a way to control the streaming quality.

```csharp
RTCRtpSendParametersRTCRtpSender.GetParameters()
RTCErrorType RTCRtpSender.SetParameters(RTCRtpSendParameters parameter)
```

Javascript API reference
- [`RTCRtpSender.getParameters`](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getParameters)
- [`RTCRtpSender.setParameters`](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/setParameters)

`RTCRtpSendParameters` contains an array of `RTCRtpEncodingParameters`.
`RTCRtpEncodingParameters` contains parameters to control video quality below.

- `active`
- `maxBitrate`
- `scaleResolutionDownBy`
- `maxFramerate`

## `active`
This flag can change on/off the streaming.

## `scaleResolutionDownBy`
This parameter is working correctly with only the software encoder.
The factor makes possible to change the frame scale.
image

## `maxBitrate`
This parameter can change the bitrate dynamically.
Add a note, this value doesn't reflect the result immediately because of the bit rate estimation.
image

## `maxFramerate`
This parameter is not implemented.


Runtime/Scripts/RTPParameters.cs
```diff
@@ -1,22 +1,107 @@
...
+        {
+            var instance = Marshal.PtrToStructure<RTCRtpSendParametersInternal>(ptr);
+            Marshal.FreeCoTaskMem(instance.encodings);
+            Marshal.FreeCoTaskMem(ptr);
+        }
     }
 
-    internal struct RTCRtpCodecParameters
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct RTCRtpSendParametersInternal
```
sindharta, line 84: I think it's better if we have this structure with the same name as the native one.

For the managed version, I think we can use another name,. such as `ManagedRTCRtpSendParameters`, or something.

karasusan: I agree with your suggestion that we have this structure in the managed code with the same name as the native one.

I am considering the name of the class in managed codes, this is the API for developers, the name is the same as the [javascript one](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSendParameters). I think this is easier to understand.
https://github.com/Unity-Technologies/com.unity.webrtc/blob/a40fcde6349bb45a631f0ce4d6924e5802cc6878/Runtime/Scripts/RTPParameters.cs#L42

I guess we can change the name of the structure in the native code and the managed one, 
`RTCRtpSendParametersStruct` or `RTCRtpSendParametersData`.

https://github.com/Unity-Technologies/com.unity.webrtc/blob/a40fcde6349bb45a631f0ce4d6924e5802cc6878/Plugin~/WebRTCPlugin/WebRTCPlugin.cpp#L468-L473

PR #119, opened 2020-05-20
feat: GetStat api

karasusan:
GetStat API provides statistics information about WebRTC features.
Made a sample to explain how to use the APIs.

030e91095dde2a72ed2f76cee2d5f715

### TODO
Sometimes the unit-test is failed on MacOS but the error is not caused by these changes.


Runtime/Scripts/WebRTC.cs
```diff
@@ -48,18 +48,34 @@ public RTCDataChannelInit(bool reliable)
...
+        {
+            return peer.GetStats(this);
         }
     }
 
     public class RTCRtpSender
     {
         internal IntPtr self;
-        internal RTCRtpSender(IntPtr ptr)
+        private RTCPeerConnection peer;
```
sindharta, line 68: This class seems to be the same as `RTCRtpReceiver.`
Maybe we can combine them ?

Runtime/Scripts/RTCStats.cs
```diff
@@ -0,0 +1,809 @@
...
+        {
+            return m_dictStats.ContainsKey(key);
+        }
+
+        public bool TryGetValue(RTCStatsType key, out RTCStats value)
+        {
+            return m_dictStats.TryGetValue(key, out value);
+        }
+
+        public IEnumerable<RTCStatsType> Keys
```
sindharta, line 768: I think we will have less code if we can provide a function that returns `m_dictStats` as `IDictionary`, so that the caller can call `Keys`, `Values`, etc

Runtime/Scripts/RTCStats.cs
```diff
@@ -0,0 +1,809 @@
...
+            if (!m_members.TryGetValue(key, out member))
+            {
+                value = null;
+                return false;
+            }
+            value = member.GetValue();
+            return true;
+        }
+
+        public IEnumerable<string> Keys
```
sindharta, line 185: I think we will have less code if we can provide a function that returns m_members as IDictionary, so that the caller can call Keys, Values, etc

Runtime/Scripts/IntPtrExtension.cs
```diff
@@ -24,5 +24,132 @@ public static string AsAnsiStringWithoutFreeMem(this IntPtr ptr)
             }
             return Marshal.PtrToStringAnsi(ptr);
         }
+
+        public static bool[] AsBoolArray(this IntPtr ptr, int length)
+        {
+            byte[] array = new byte[length];
+            Marshal.Copy(ptr, array, 0, length);
+            Marshal.FreeCoTaskMem(ptr);
+            return Array.ConvertAll(array, Convert.ToBoolean);
```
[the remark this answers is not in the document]
karasusan: Unfortunately, `Marshal.Copy` method can not set generic arguments.

```
        public static void Copy(IntPtr source, int[] destination, int startIndex, int length);
        public static void Copy(IntPtr source, double[] destination, int startIndex, int length);
        public static void Copy(IntPtr source, float[] destination, int startIndex, int length);
        public static void Copy(IntPtr source, long[] destination, int startIndex, int length);
        public static void Copy(IntPtr source, short[] destination, int startIndex, int length);
        public static void Copy(IntPtr source, char[] destination, int startIndex, int length);
        public static void Copy(IntPtr[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(IntPtr source, byte[] destination, int startIndex, int length);
        public static void Copy(byte[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(float[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(long[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(short[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(char[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(int[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(double[] source, int startIndex, IntPtr destination, int length);
        public static void Copy(IntPtr source, IntPtr[] destination, int startIndex, int length);
```

Runtime/Scripts/IntPtrExtension.cs
```diff
@@ -24,5 +24,132 @@ public static string AsAnsiStringWithoutFreeMem(this IntPtr ptr)
...
+        }
+
+        public static int[] AsIntArray(this IntPtr ptr, int length)
```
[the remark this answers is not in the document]
karasusan: https://github.com/Unity-Technologies/com.unity.webrtc/pull/119#discussion_r433041625

Runtime/Scripts/WebRTC.cs
karasusan: This pull request has the changes for `RTCRtpSender` class.
It could be better as it is.
https://github.com/Unity-Technologies/com.unity.webrtc/pull/118/files#diff-7ef9c5b7b61c1912384f936f85cc13dcR60-R94

PR #123, opened 2020-06-03
fix: Memory leak when using RTX sample

karasusan:
## This PR
This PR is for fixing the memory leak issue #123 reported by a user on the Unity Forum.

This [line](https://github.com/Unity-Technologies/com.unity.webrtc/pull/123/files#diff-9b61b8ef15be0cd78f836c22b23ffd33R117) is needed to make the GPU command list.

```c++
ThrowIfFailed(m_commandAllocator->Reset());
```

To fix this issue, fixed the following point.

- Add native test of DX12
- Check HRESULT with `ThrowIfFailed` method

I referenced [this site]() to know how to use DX12 API.


sindharta:
Does this PR fix the memory leak ?
I didn't see anything that seems to solve the issue.

PR #130, opened 2020-06-26
Check the correct variable to see if it exists.

tmiv:


karasusan:
Oh, this is shameful...
Thank you for  your cooperation!

PR #131, opened 2020-07-02
fix: stats review

kannan-xiao4:
Fix:
- Move SaveButton to toolbar
- If can't save file, SaveIcon disable.
- Change menu name (WebRTCInternals -> WebRTC Stats)
- If any stats don't display, display notice.
- if input value 1000 over, change grip unit.

Problem : don't displayed  anything when it is selected AudioSource.
- fix on https://github.com/Unity-Technologies/com.unity.webrtc/pull/125 (commit: 77ba287f8b372a53d2281993848d0840284371fb)

Light
image

Dark
image


Editor/GraphView.cs
```diff
@@ -64,6 +65,24 @@ private static void Draw(Rect area, ref List<KeyValuePair<DateTime, float>> data
...
+                current /= 1000;
+                unitCount *= 1000;
+                unitIndex++;
+            }
+
+            if (unitIndex > unitStr.Length - 1)
+            {
+                unitIndex = unitStr.Length - 1;
+            }
+
```
karasusan, line 85: It might be better this routine is static method.

Editor/PeerListView.cs
```diff
@@ -30,7 +32,12 @@ public VisualElement Create()
                     var button = new Button(() =>
                     {
                         OnChangePeer?.Invoke(peerConnection);
-                    }) {text = $"peer {peerConnection.GetHashCode()}"};
+                    }) {text = $"peer {peerConnection.GetHashCode()}",};
+
+                    if (EditorGUIUtility.isProSkin)
+                    {
+                        button.style.backgroundColor = new Color(70 / 255f, 70 / 255f, 70 / 255f);
```
karasusan, line 39: The instance of Color can be changed readonly variable.

Editor/WebRTCStats.cs
```diff
@@ -36,7 +38,20 @@ public static void Show()
         private void OnEnable()
         {
             var root = this.rootVisualElement;
-            root.Add(new Button(() =>
+            root.style.backgroundColor = EditorGUIUtility.isProSkin ? new Color(45/255f,45/255f,45/255f) : Color.white;
```
karasusan, line 41: The instance of Color can be changed readonly variable.

Editor/GraphView.cs
kannan-xiao4: This method (Draw) is static method.

Does that mean I should cut out function?

karasusan: Yes, it make the operation clarify.

Issue #138, opened 2020-07-14
[BUG] MediaStreamSample doesn't work


iiiyx:
**Describe the bug**
MediaStreamSample doesn't work. It stucks on setting peer description. 

**To Reproduce**
Just run the sample and add tracks
After that, the program will stuck on setting local peer description: `pc1.SetLocalDescription(...)`

**Expected behavior**
WebRTC connection established and media flows

**Environment (please complete the following information):**
 - OS: Windows 10 x64
 - Unity Version 2019.4.1f1
 - Package version 2.0.4-preview

**Additional context**
The problem appears when there are tracks in the stream. I.e. PeerConnectionSample works because there are no tracks.

karasusan:
We reproduced the issue and will fix it for the next version. Thanks!

karasusan:
fixed this issue with #161.

PR #161, opened 2020-08-20
fix: media stream sample

kannan-xiao4:
issue: https://github.com/Unity-Technologies/com.unity.webrtc/issues/138

fix point only this commit https://github.com/Unity-Technologies/com.unity.webrtc/commit/4e659ff2ac961611cfca1c0baa71634970620036
- need enable video offer option falg

others
- add get signaling state method 
- add sdp type about rollback

### DIFF
--- a/Runtime/Scripts/RTCPeerConnection.cs
+++ b/Runtime/Scripts/RTCPeerConnection.cs
@@ -117,7 +117,7 @@ namespace Unity.WebRTC
         /// var signalingState = peerConnection.SignalingState;
         /// </code>
         /// </example>
-        /// <seealso cref="SignalingState"/>
+        /// <seealso cref="ConnectionState"/>
         public RTCSignalingState SignalingState
         {
             get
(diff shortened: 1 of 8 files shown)


sindharta (approved):

Runtime/Scripts/RTCPeerConnection.cs
```diff
@@ -106,6 +106,26 @@ public RTCPeerConnectionState ConnectionState
...
+        /// the current state of the peer connection by returning one of the
+        /// <see cref="RTCSignalingState"/> enum.
+        /// </summary>
+        /// <example>
+        /// <code>
+        /// var peerConnection = new RTCPeerConnection(configuration);
+        /// var signalingState = peerConnection.SignalingState;
+        /// </code>
+        /// </example>
+        /// <seealso cref="SignalingState"/>
```
karasusan, line 120: `seealso` is pointing itself. 🤔 
`RTCSignalingState` is correct?

kannan-xiao4: I was wrong about the `seealso` tag.
I think it's the `connectionState` that should be referenced (it's related this).
So I'll change it.

Issue #100, opened 2020-04-22
Video stream remains with bad quality after a short network degradation


eirikwah:
When the bandwidth between server and client is degraded from 20 Mbps to 2 Mbps, the video stream is degraded, and this is expected and works fine. However, when the network is back at 20 Mbps again (for example after 20 seconds), the video stream is not improved even if I wait several minutes. I need to disconnect the client and reconnect to get good video quality again.

Is there a setting/code I should use to fix this, or is it a bug or missing feature?

---
More info on my test setup:

I am testing on Windows x64, running Unity 2019.3.10f1, and Unity Render Stream 2.0.0-preview.4, and thus WebRTC 2.0.0-preview.3. I use HW encoding on the server.

I got the same result on Unity Render Stream version 1.2.3-preview and WebRTC 1.x (or, actually much worse because the HW encoding on bad network gave artefacts that thankfully has been resolved in later versions).

Using SW encoding seems to improve this issue, but it is hard to find a consistent pattern.

I have a very controlled environment for testing and for the bandwidth: The server and client are on the same LAN, and the client runs on Chrome on macOS and I use "Network Link Conditioner" to control the bandwidth on the mac.

Degraded video:
![image](https://user-images.githubusercontent.com/1539783/79985500-fbadce80-84aa-11ea-806a-bf07eb0fdea0.png)

Normal video:
![image](https://user-images.githubusercontent.com/1539783/79985629-2ac44000-84ab-11ea-8c41-b6d32a9118eb.png)

Thank you for this great library! We are looking forward to the next version(s)! :-)

karasusan:
I appreciate your report in detail.
OK, I am tackling it.

karasusan:
As you say, I found strange results. It might be wrong in the implementation.
I am digging into the native code.

## Sofware encoder
<img width="385" alt="Screen_Shot_2020-04-23_at_3_40_27_PM" src="https://user-images.githubusercontent.com/1132081/80067903-a68cbe00-8579-11ea-8daf-98ffdf4ca6d6.png">

## Hardware encoder
<img width="444" alt="Screen_Shot_2020-04-23_at_2_53_15_PM" src="https://user-images.githubusercontent.com/1132081/80068005-d20fa880-8579-11ea-9676-f6fa4facdf97.png">

eirikwah:
Thank you very much for looking into this. Let me know if I can do anything to help (for example testing).

eirikwah:
[...]

PR #162, opened 2020-08-20
fix: Video stream remains with bad quality after a short network degradation

karasusan:
Related issue https://github.com/Unity-Technologies/com.unity.webrtc/issues/100 .

### Problem

This issue is the feature of recovering video quality do not work well after low bandwidth.
More detail of the issue is in the thread (#100).

### The cause of issue

The argument of `VideoEncoder::SetRates` has a `framerate_fps` parameter which is a value estimating frame rate.
This parameter is very high compared real interval of each encoding process. So NvEnc outputs the encoded data that data size smaller than expected. The estimation of bandwidth keeps low unless the encoder sends the proper sized data.

### Fixes

To fix this issue, need the individual way of estimating frame rate. `webrtc::RateStatistics` class was used for the implementation.

### TODO

The present implementation of the encoder is messy. In the next version 2.2, we will refine it.

### DIFF
--- a/Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoder.cpp
+++ b/Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoder.cpp
@@ -241,18 +241,6 @@ namespace webrtc
     constexpr double kLowRateFactor = 1.0;
     constexpr double kHighRateFactor = 2.0;
 
-    uint32_t Interpolate(uint32_t low,
-        uint32_t high,
-        double bandwidth_headroom_factor) {
-
-        RTC_DCHECK_GE(bandwidth_headroom_factor, kLowRateFactor);
-        RTC_DCHECK_LE(bandwidth_headroom_factor, kHighRateFactor);
-
-        // |factor| is between 0.0 and 1.0.
-        const double factor = bandwidth_headroom_factor - kLowRateFactor;
-
-        return static_cast<uint32_t>(((1.0 - factor) * low) + (factor * high) + 0.5);
-    }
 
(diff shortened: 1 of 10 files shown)


Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoder.cpp
```diff
@@ -85,7 +89,10 @@ namespace webrtc
         std::memcpy(&nvEncConfig, &presetConfig.presetCfg, sizeof(NV_ENC_CONFIG));
         nvEncConfig.profileGUID = NV_ENC_H264_PROFILE_BASELINE_GUID;
         nvEncConfig.gopLength = nvEncInitializeParams.frameRateNum;
-        nvEncConfig.rcParams.averageBitRate = m_bitrateAdjuster->GetAdjustedBitrateBps();
+        nvEncConfig.rcParams.rateControlMode = NV_ENC_PARAMS_RC_CBR_LOWDELAY_HQ;
+        nvEncConfig.rcParams.averageBitRate = (static_cast<unsigned int>(5.0f *
+            nvEncInitializeParams.encodeWidth *
+            nvEncInitializeParams.encodeHeight) / (1280 * 720)) * 100000;
```
kannan-xiao4, line 95: this value `(1280 * 720)`, could you use this filed value?

```suggestion
            nvEncInitializeParams.encodeHeight) / (m_width* m_height)) * 100000;
```

karasusan: Oh, I will fix it

kannan-xiao4 (approved):

PR #194, opened 2020-09-24
chore: fix rtcerror message handling

kannan-xiao4:
Changed to use primitive elements instead of RTCError.
I was going to take it as Struct, but it didn't work out...


Plugin~/WebRTCPlugin/WebRTCPlugin.h
```diff
@@ -110,15 +109,6 @@ namespace webrtc
         Video
     };
 
-    struct RTCError
```
kannan-xiao4, line 113: this struct is not used

karasusan (approved):

Issue #151, opened 2020-08-03
[BUG] Crash when starting sample scene in linux


GetJon:
**Describe the bug**
A clean project started from the Render Streaming HDRP template crashes when pressing play on Linux. 

I suspect either a missing library: `Fallback handler could not load library /home/andrewsj/Unity/Hub/Editor/2020.1.0f1/Editor/Data/Mono/lib/libPackages/com.unity.webrtc/Runtime/Plugins/x86_64/libwebrtc.so` or insufficient official installation documentation causing me to miss out vital steps.

**To Reproduce**
Following installation instructions from: https://gamefromscratch.com/unity-release-render-streaming-over-webrtc/ for lack of anything better.
1. Go to https://github.com/Unity-Technologies/UnityRenderStreaming/releases, download latest releases
1. Go to https://github.com/Unity-Technologies/com.unity.webrtc/releases, download latest releases
1. Copy webrtc-linux.zip to /home/.../Unity/Hub/Editor/2020.1.0f1/Editor/Data/Resources/PackageManager/Editor
1. Copy com.unity.template.renderstreaming-hd-2.0.2-preview.tgz to /home/.../Unity/Hub/Editor/2020.1.0f1/Editor/Data/Resources/PackageManager/ProjectTemplates
1. Start webserver application in terminal
1. Create new unity project from the Render Streaming HDRP template
1. Enter the webserver IP and port number to the RenderStreaming gameobject's URL Signalling variable
1. Press play on the editor
1. Unity crashes

**Expected behavior**
Unity not to crash, and for the sample project to run successfully.

**Log file**
[Editor.log](https://github.com/Unity-Technologies/com.unity.webrtc/files/5015555/Editor.log)

**Crash extract**
```
Reloading assemblies for play mode.
Begin MonoManager ReloadAssembly
./Editor/Platform/Linux/UsbDevices.cpp:UsbDevicesQuery
Symbol file LoadedFromMemory doesn't match image /home/andrewsj/TT-dev/unity-projects/webrtc2/Library/PackageCache/com.unity.ext.nunit@1.0.0/net35/unity-custom/nunit.framework.dll
Native extension for LinuxStandalone target not found
Native extension for WebGL target not found
Refreshing native plugins compatible for Editor in 0.53 ms, found 1 plugins.
Preloading 1 native plugins for Editor in 0.08 ms.
[MODES] Loading mode Default (0) for mode-current-id-webrtc2
[...]

PR #198, opened 2020-09-28
feat: vulkan support

karasusan:
This change is Vulkan API support on Windows and Linux platform. Therefore, you can use the video streaming with HDRP projects on Linux.

## CUDA interop with Vulkan API
To support hardware encoding with Vulkan API, need to convert the render texture provided by Unity to CUDA resources. We need to add vulkan extensions when creating the vulkan device to use the interop between CUDA and Vulkan API.
Unity provides the way to add extensions as a native API `InterceptInitialization` function.
You can see how to use this API on the GitHub repository.
https://github.com/Unity-Technologies/NativeRenderingPlugin/blob/master/PluginSource/source/RenderAPI_Vulkan.cpp

### List of Vulkan extensions used by native plugin

- VK_KHR_GET_PHYSICAL_DEVICE_PROPERTIES_2_EXTENSION_NAME
- VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME
- VK_KHR_EXTERNAL_SEMAPHORE_CAPABILITIES_EXTENSION_NAME

- VK_KHR_EXTERNAL_MEMORY_EXTENSION_NAME
- VK_KHR_EXTERNAL_SEMAPHORE_EXTENSION_NAME
- VK_KHR_EXTERNAL_MEMORY_FD_EXTENSION_NAME
- VK_KHR_EXTERNAL_SEMAPHORE_FD_EXTENSION_NAME
- VK_KHR_EXTERNAL_MEMORY_WIN32_EXTENSION_NAME
- VK_KHR_EXTERNAL_SEMAPHORE_WIN32_EXTENSION_NAME

## TODO
Software encoder is poor performance than other graphics APIs.


Plugin~/WebRTCPlugin/Context.cpp
```diff
@@ -253,8 +253,9 @@ namespace webrtc
     bool Context::EncodeFrame(MediaStreamTrackInterface* track)
     {
         UnityVideoTrackSource* source = GetVideoSource(track);
-        if(source != nullptr)
-            source->OnFrameCaptured();
+        if (source != nullptr)
+            return false;
+        source->OnFrameCaptured();
```
sindharta-tanuwijaya, line 258: source is nullptr here ?

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
```diff
@@ -143,28 +160,56 @@ bool VulkanGraphicsDevice::CopyResourceV(ITexture2D* dest, ITexture2D* src) {
...
+            kUnityVulkanResourceAccess_PipelineBarrier,
+            &unityVulkanImage))
+        {
+            return false;
+        }
+        image = unityVulkanImage.image;
     }
+    else
+    {
+        UnityVulkanImage* unityVulkanImage = static_cast<UnityVulkanImage*>(nativeTexturePtr);
```
sindharta-tanuwijaya, line 189: The method clearly says CopyResourceFromNativeV, so nativeTexturePtr should be a native texture pointer, not a pointer to UnityVulkanImage.
I don't think we should make the code more complicated, just because we want to test it.
And because there is a branching here, we are not testing the actual runtime code anyway

Is is possible to pass the native texture directly from the testing code instead ?

Plugin~/WebRTCPlugin/Context.cpp
karasusan: Oh, I mistaken merge when rebase ..

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
karasusan: Yeah, you are right. I am also sure that we should not make this branching.
But the type of a native texture pointer is not public from Unity, so we cannot create the texture without Unity.

As an alternative idea, that is defining the texture pointer of argument of this method as `UnityVulkanImage`.
How about this idea?

Plugin~/WebRTCPlugin/Context.cpp
karasusan: fixed

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
sindharta-tanuwijaya: But  you want to keep the name as CopyResourceFromNativeV(), or do you want to create a new method ?

sindharta-tanuwijaya: Hm, not ideal and might be misleading, but I see that this might be a good trade off rather than changing the structure of the code.
Ok.

Plugin~/WebRTCPlugin/UnityRenderEvent.cpp
```diff
@@ -64,10 +83,17 @@ static void UNITY_INTERFACE_API OnGraphicsDeviceEvent(UnityGfxDeviceEventType ev
     case kUnityGfxDeviceEventInitialize:
     {
         s_mapEncoder.clear();
+
+        IGraphicsDevice* device = GetGraphicsDevice();
+        if (device != nullptr)
+            device->InitV();
```
kannan-xiao4, line 89: `device->InitV()` is calld in `GetGraphicsDevice()`. 
https://github.com/Unity-Technologies/com.unity.webrtc/blob/feat/vulkan/Plugin~/WebRTCPlugin/GraphicsDevice/GraphicsDevice.cpp#L130
so,It is called twice here,. is this no problem ?

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
karasusan: fixed

kannan-xiao4 (approved):

sindharta-tanuwijaya (approved):

karasusan:
https://github.com/Unity-Technologies/com.unity.webrtc/issues/151

PR #200, opened 2020-09-29
fix: fixed crash when using debug build plugin

karasusan:
This pull request fixes bugs occurred when using debug build plugin.

- Add command to build `googletest` for debug
- Manage a list of references of `MediaStreamTrackInterface` to check the track whether disposed or not
- Add a mutex to `UnityVideoTrackSource` to avoid access from render thread after dispose it


kannan-xiao4 (approved):
lgtm, I noticed that I need a Context mutex in RenderEvent, thanks

PR #208, opened 2020-10-07
Add a nullptr check for profiler in development build

sipsorcery:
When doing some testing on the native plugin I had closed the profiler and started getting null pointer exceptions.


karasusan:
Thanks for your contribution!

karasusan (approved):

PR #211, opened 2020-10-08
Removes bitrate parameter from CameraExtension capture methods

sipsorcery:
* The bitrate parameter is currently mandatory but is not used. It's also an unusual parameter to require at the application layer since the WebRTC plumbing typically employs a dynamic bit rate based on RTCP feedback.

* If it is intended to give the application a level of control over the encoder parameters it would probably make more sense to allow the frame rate to be set and/or employ an enum with values like "real-time", "best-quality" etc.


sipsorcery:
I noticed there is a native bit-rate property that can be set on the NVenc encoder so I guess that's what the managed properties are eventually intended for,

Issue #204, opened 2020-10-03
[REQUEST] Direct access to Nvenc encoding


sipsorcery:
Is there any chance the Unity webrtc package could expose direct access to the NVenc video encoder? The idea being that other WebRTC and VoIP implementations could then take advantage of it.

Some WebRTC use cases require different signalling mechanisms (or in the case of VoIP completely different protocols) or for differences to the way `libwebrtc` does things (e.g. enable private IP addresses). Other .NET based implementations can be used in Unity right now but the video, and to a lesser extent, audio encoding is never going to be that great without the hardware encoders.

Ideally a C# API would be nice but failing that even a documented, supported C++ API that could be used with PInvoke would be good.

rafcsoares:
Totally Agree!
Exposing NVec|VP8|VP9|H264 Encoder|Decoders can benefit other projects not directly related to Unity's WebRTC implementation.

karasusan:
@racsoares
@sipsorcery
My understanding that providing managed API for encoding/decoding using HWA, is it right?

sipsorcery:
> My understanding that providing managed API for encoding/decoding using HWA, is it right?

Yes, that would be very helpful.

karasusan:
@sipsorcery
I understood. This idea is great.
I guess creating the different package for HWA codec would be better.
I will discuss about the idea with my team. Thanks.

haoqc86:
> @sipsorcery I understood. This idea is great. I guess creating the different package for HWA codec would be better. I will discuss about the idea with my team. Thanks.

It will be very useful~

PR #217, opened 2020-10-12
Add callback to native video source for encoded frames 

sipsorcery:
This PR is an attempt at implementing the feature request from #204.

It adds the ability to set a callback on the native `UnityVideoSource` class so managed scripts can get video frames after the post-capture processing.

**TLDR; Currently looking for feedback as to whether the approach in this PR will be considered and thus whether reamingin issues should be solved and more testing carried out.**

Initial testing indicates this approach or something similar is very worthwhile. A comparison was done between capturing video frames from a camera purely in a C# and contrasted with the native callback in this PR. The difference in frame rates was over 5x, approx 15fps compared to approx 80 fps. The test was done at 480p on Window 10.

The Unity WebRTC plugins use of a native graphic device texture for video capture makes the difference between an acceptable streaming experience and a sub-standard one. If something along the lines of this hook can be incorporated into this plugin it will open up video streaming from Unity camera textures to managed script streaming applications such as WebRTC, RTSP, SIP etc.

This PR is flagged as Work In Progress as their are outstanding issues that could take a large chunk of time to solve and I'd like to get some initial feedback before expending that time. The main issues are:

 - The editor freezes after the initial execution. It seems to be a problem with the managed assembly unload not completing cleanly.

 - Testing so far has only been with software encoding and processing I420 buffers which are then handed off to `libvpx` for encoding. The next step is to attempt to handle the H264 encoded samples received when NVenc hardware encoding is used and stream those directly.

Screenshot of native callback and the editor running at approx 80fps (the frame rate is double checked with Chrome's webrtc diagnostics page). Note the screenshot shows 63fps, my screen capture must have caused it to drop from around 80.

unity_videosource_withwebrtcplugin

Screenshot with managed only capture and editor running at approx 15 fps.

unity_videosource_noplugin

### DIFF
--- a/Plugin~/WebRTCPlugin/UnityVideoTrackSource.cpp
+++ b/Plugin~/WebRTCPlugin/UnityVideoTrackSource.cpp
@@ -106,6 +106,8 @@ void UnityVideoTrackSource::DelegateOnFrame(const ::webrtc::VideoFrame& frame) {
 
             auto i420Buffer = frame.video_frame_buffer()->ToI420();
             if (i420Buffer != nullptr) {
+                // If the WebRTC context was initialised with software encoding then the post frame capture
+                // will result in I420 buffer samples.
                 auto bufferSz = ::webrtc::CalcBufferSize(::webrtc::VideoType::kI420, width, height);
                 static std::vector<uint8_t> buffer(bufferSz);
                 int extractSz = ::webrtc::ExtractBuffer(i420Buffer, bufferSz, buffer.data());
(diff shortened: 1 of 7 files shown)


karasusan:
Hi, I guess your request is to provide API to use encoded frames with the managed script. Is it right?
It might be able to realize this feature using "Insertable Streams API".
https://w3c.github.io/webrtc-insertable-streams/

If many developers wish it, we could prioritize to develop it.

unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.webrtc?pullRequest=217) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.webrtc?pullRequest=217) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.webrtc?pullRequest=217) it.</sub>

sipsorcery:
> It might be able to realize this feature using "Insertable Streams API".

That would be close to what I'm describing but the idea is to provide the managed script hook before going into the `libwebrtc` pipeline, i.e. no `RTCPeerConnection` instance required.

unity_webrtc-hooks

In the above diagram the `Network/Transport` block is currently fulfilled by `libwebrtc` and the `Signaling` block by `Render Streaming` or an alternative mechanism such as node or web sockets.

If the managed hooks are added in the locations shown in the diagram and as implemented in this PR then the both the `Network/Transport` and `Signaling` blocks can be replaced by different implementations or protocols. For example with this PR I can use my [C# WebRTC](https://github.com/sipsorcery/sipsorcery) library instead of the `libwebrtc` infrastructure. That has pros and cons for different use cases. Or with the same library I can use `SIP` and `RTP` to place a VoIP call with the video stream from a Unity Camera. This can currently be done without the `Unity WebRTC` plugin but the frame rates are way too low without being able to use the graphics devices for the frame capture.

For a very small addition to the API in this plugin it can open up a realm of real-time communications possibilities rather than being completely coupled to `libwebrtc`.

sipsorcery:
Access to raw media streams is on the roadmap for WebRTC via the [MediaStream Processing API](https://dvcs.w3.org/hg/audio/raw-file/tip/streams/StreamProcessing.html). That sort of API is exactly the goal of this PR.

karasusan:
> Access to raw media streams is on the roadmap for WebRTC via the [MediaStream Processing API](https://dvcs.w3.org/hg/audio/raw-file/tip/streams/StreamProcessing.html). That sort of API is exactly the goal of this PR.

Yes, I agree with your suggestion.
I created the ticket for resolving this issue. If you have more opinions, you can add comments on this ticket.
Would you mind if I closed this PR?

PR #245, opened 2020-10-28
chore: Organize CI jobs to test package on each platforms without GPU

karasusan:
Up until this change, package testing are ran on each platforms with GPU. 
This PR adds jobs that run test on the system not depend on GPU.

note:
CI system kicks a trigger jobs for testing with GPU as in the past.


.yamato/upm-ci-webrtc-packages.yml
```diff
@@ -143,27 +205,50 @@ test_{{ package.name }}_{{ param.platform }}_{{ param.backend }}_{{ platform.nam
...
+    {% endfor %}
+
+test_{{ package.name }}_gpu:
+  name : Test {{ package.packagename }} all platforms with GPU
+  triggers:
+    branches:
+      only:
+      - "/.*/"
+      except:
+      - "master"
```
kannan-xiao4, line 238: I think this change for avoiding use `gpu vm`.
so, `tiggers` is defined on `test_{{ package.name}}_nogpu`, isn't it?

sindharta-tanuwijaya (approved):

.yamato/upm-ci-webrtc-packages.yml
karasusan: In the end, we should do it, but it has not done yet.
Currently, the test would be failed without GPU.

kannan-xiao4 (approved):

PR #227, opened 2020-10-16
Fix: not working on MacOS using color space gamma

kannan-xiao4:
Issue: https://github.com/Unity-Technologies/com.unity.webrtc/issues/90 

Important change is this file changes : https://github.com/Unity-Technologies/com.unity.webrtc/compare/fix/work-using-colorspacegamma-onmac?expand=1#diff-cd6d583da8e66fb51ab121850c37e94b333726be46d5099d2ba57b4d4670b0d0R36

- I add `colorspace` argument on `CreateContext` Method.

PR #253, opened 2020-11-10
fix: Fix crash bug when using vulkan API when the application ended

karasusan:
This PR fixes the crash bug that occurs during terminate the application using vulkan API.
It might be the cause of crash is the `vkDestroyDevice` function called from the plugin side. This logical device is allocated from Unity side.

### DIFF
--- a/Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
+++ b/Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
@@ -55,9 +55,8 @@ bool VulkanGraphicsDevice::InitV() {
 //---------------------------------------------------------------------------------------------------------------------
 
 void VulkanGraphicsDevice::ShutdownV() {
-    VULKAN_SAFE_DESTROY_COMMAND_POOL(m_device, m_commandPool, m_allocator);
-    vkDestroyDevice(m_device, NULL);
     m_cudaContext.Shutdown();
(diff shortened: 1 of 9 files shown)


Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
```diff
@@ -55,9 +55,8 @@ bool VulkanGraphicsDevice::InitV() {
 //---------------------------------------------------------------------------------------------------------------------
 
 void VulkanGraphicsDevice::ShutdownV() {
-    VULKAN_SAFE_DESTROY_COMMAND_POOL(m_device, m_commandPool, m_allocator);
-    vkDestroyDevice(m_device, NULL);
```
kannan-xiao4, line 59: This is avoiding test failed on vulkan. https://github.com/Unity-Technologies/com.unity.webrtc/pull/227/commits/492204bbd29252e8cf73078afd70a9b23b6ad5f7
Is plugin native test passed?

karasusan: OK, I will check the native test.

karasusan: All native tests are passed on Windows.
Which platform can we reproduce the bug?

kannan-xiao4 (approved):

PR #323, opened 2021-02-25
Fix: crash when exit Play mode

kannan-xiao4:
Some sample occurs crash when exit Play mode after press `HangUp`.
It caused by `VideoTrack` does not dispose。


kannan-xiao4:
change fix to https://github.com/Unity-Technologies/com.unity.webrtc/pull/324

PR #345, opened 2021-03-24
feat: Add sample menu scene

karasusan:
There are 11 sample scenes in this package.
We had to make a standalone build many times manually to test samples on specific platforms.
This fix provides a menu scene that can jump each scene to test them.

The button placed top-right is to back the menu scene.
Mar-24-2021 17-17-23


kannan-xiao4 (approved):
LGTM!

PR #367, opened 2021-04-16
chore: remove offer to receive video/audio options

kannan-xiao4:
from `peer_connection_interface.h`
```
    // These options are left as backwards compatibility for clients who need
    // "Plan B" semantics. Clients who have switched to "Unified Plan" semantics
    // should use the RtpTransceiver API (AddTransceiver) instead.
    //
    // offer_to_receive_X set to 1 will cause a media description to be
    // generated in the offer, even if no tracks of that type have been added.
    // Values greater than 1 are treated the same.
    //
    // If set to 0, the generated directional attribute will not include the
    // "recv" direction (meaning it will be "sendonly" or "inactive".
```

So, I use Direction property on Transceiver for changing `Send / Recv` .

### DIFF
--- a/Plugin~/WebRTCPlugin/PeerConnectionObject.cpp
+++ b/Plugin~/WebRTCPlugin/PeerConnectionObject.cpp
@@ -240,6 +240,7 @@ namespace webrtc
     {
         webrtc::PeerConnectionInterface::RTCOfferAnswerOptions _options;
         _options.ice_restart = options.iceRestart;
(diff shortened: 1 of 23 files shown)


karasusan (approved):

Logicator:
The removal of the options to send audio or video breaks my use case for this package. The implementation of our signaling server (mediasoup) relies on 1-way webRTC connections with clients being either sendonly or reconly. With audio/video options on both ends. Is is possible to undo this and give the options back? Thank you and thank you and your team for working on this package.

PR #382, opened 2021-04-26
chore: Change GitHub cli for CI

karasusan:


.yamato/upm-ci-publish-github-release.yml
```diff
@@ -16,28 +16,42 @@ publish_com.unity.webrtc_github_release:
   name: Publish com.unity.webrtc to Github Release
   agent:
     type: Unity::VM
-    image: package-ci/win10:stable
+    image: package-ci/ubuntu:stable
     flavor: b1.large
+  variables:
+    GIT_TAG: Foo
```
kannan-xiao4, line 22: Is this variables needed ?

kannan-xiao4: I see, `GIT_TAG` value is `Foo`, is this correct?
I think `GIT_TAG` value need to set webrtc-version string (example `2.4.0-exp.1`).

karasusan: Usually, we need to set the parameter on Yamato UI.
We don't have to change this line for every release.
image

karasusan: OK, `Foo` may not suitable for us, so how about `You need to change this parameter on Yamato UI`?
Is it more clear than `Foo`?

karasusan: > 
```suggestion
  variables:
    GIT_TAG: You need to change this parameter on Yamato UI
```

kannan-xiao4 (approved):

PR #440, opened 2021-06-16
feat: Refine AudioStreamTrack class

karasusan:
This pull request improves the audio streaming features.

- Support multiple streaming tracks concurrently
- `AudioSource` instances can be used for streaming sources

I have a plan for other updates below.

- Add the `Audio` sample which uses the microphone as a streaming source
- Add the new sample to test streaming stereo audio
- Support audio receiver


Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
```diff
@@ -6,25 +6,5 @@ namespace unity
 namespace webrtc
 {
 
-    void DummyAudioDevice::ProcessAudioData(const float* data, int32 size)
```
kannan-xiao4, line 9: I think no need this file.

kannan-xiao4 (approved):

Issue #441, opened 2021-06-17
[BUG] Native plugin compiling fails on Windows (webrtc.lib error LNK2019: unresolved external symbol __std_reverse_copy_trivially_copyable_4)


rollersteaam:
**Describe the bug**
Running `build_plugin_win.cmd` fails with a linker error. It might not have been able to link a dependency.

**To Reproduce**
1. Get a fresh clone of the project on a Windows 10 machine
2. Run `build_plugin_win.cmd`

**Expected behavior**
Building succeeds.

**Screenshots**
![image](https://user-images.githubusercontent.com/6756697/122464957-5c517200-cfaf-11eb-88f1-752d80284780.png)

**Environment (please complete the following information):**
 - OS: Windows 10 Pro 64-bit (10.0, Build 19043)
 - Package version: 2.4.0-exp.2

**Additional context**
Full error message:
```
webrtc.lib(auto_correlation.obj) : error LNK2019: unresolved external symbol __std_reverse_copy_trivially_copyable_4 re
ferenced in function "public: void __cdecl webrtc::rnn_vad::AutoCorrelationCalculator::ComputeOnPitchBuffer(class rtc::
ArrayView<float const ,432>,class rtc::ArrayView<float,147>)" (?ComputeOnPitchBuffer@AutoCorrelationCalculator@rnn_vad@
webrtc@@QEAAXV?$ArrayView@$$CBM$0BLA@@rtc@@V?$ArrayView@M$0JD@@5@@Z) [C:\Users\Jordan\Desktop\sv-for-ml\Assets\com.unit
y.webrtc.simplyvideo\Plugin~\build64\WebRTCPlugin\WebRTCPlugin.vcxproj]
C:\Users\Jordan\Desktop\sv-for-ml\Assets\com.unity.webrtc.simplyvideo\Runtime\Plugins\x86_64\webrtc.dll : fatal error L
NK1120: 1 unresolved externals [C:\Users\Jordan\Desktop\sv-for-ml\Assets\com.unity.webrtc.simplyvideo\Plugin~\build64\W
ebRTCPlugin\WebRTCPlugin.vcxproj]
```

Weirdly, opening the `webrtc.sln` file in the `build64` folder in Visual Studio 2019 and building ALL_BUILD completely succeeds. 

[...]

PR #438, opened 2021-06-14
Help Windows devs save major debugging time with first time builds

rollersteaam:
This PR relates to this issue: https://github.com/Unity-Technologies/com.unity.webrtc/issues/434

A Windows developer's first time build will fail if they have older versions of Windows SDK installed. However the C++ compiler is incredibly vague, and debugging the issue can take days to figure out. I have added a warning to `README.md` to prevent this.

PR #443, opened 2021-06-18
doc: Save Windows devs first time build debug time (issue #441)

rollersteaam:
Solves issue https://github.com/Unity-Technologies/com.unity.webrtc/issues/441.

Issue https://github.com/Unity-Technologies/com.unity.webrtc/issues/441 shows that on some Windows environments, building will fail at the linking stage when running `build_plugin_win.cmd`, probably due to having multiple C++ compilers on the user's computer.

This PR adds a message in README.md in similar vein to PR https://github.com/Unity-Technologies/com.unity.webrtc/pull/438 that gives a solution/workaround for Windows devs encountering linker issues.

I'm not really sure this is the best solution to the problem, ideally a change to the CMake that would make it link the correct standard library version would be best. But as a temporary thing, this works, and is [actually a valid use of CMake as described in my issue post.](https://github.com/Unity-Technologies/com.unity.webrtc/issues/441)

I think this might actually lead to a better developer experience too, because using the project files made by CMake when deving probably leads to better intellisense.

### DIFF
--- a/Plugin~/README.md
+++ b/Plugin~/README.md
@@ -109,6 +109,7 @@ To build plugin, you need to execute command in the `BuildScripts~` folder.
 - [BuildScripts~/build_plugin_ios.sh](../BuildScripts~/build_plugin_ios.sh)
 - [BuildScripts~/build_plugin_linux.sh](../BuildScripts~/build_plugin_linux.sh)
 - [BuildScripts~/build_plugin_win.cmd](../BuildScripts~/build_plugin_win.cmd)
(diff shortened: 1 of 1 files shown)


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.webrtc?pullRequest=443) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.webrtc?pullRequest=443) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.webrtc?pullRequest=443) it.</sub>

kannan-xiao4 (approved):

PR #453, opened 2021-07-01
feat: Audio receiver and audio sample

karasusan:
This pull request contains two main changes.
- Audio receiver support 
- Add an `Audio` sample

This demo shows the audio streaming feature. It sends the stereo audio source(Left:400Hz, Right:1600Hz) from the peer, and the other peer receives it and renders it. The audio is converted monaural.
WebRTC_ 2021-07-01 11-40-42_Trim (3) (1)

## TODO
- Stereo audio (multichannel audio) support
- Add audio codec options


kannan-xiao4 (approved):

rollersteaam:
This is really fucking cool!

Can I ask how you found out how to implement this? I could not find documentation for libwebrtc native _**anywhere**_ at all

karasusan:
@rollersteaam
There is a hint in the chromium codebase. Basically, we have to read the code if we want to understand it.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/examples/unityplugin/simple_peer_connection.cc;l=559-568

rollersteaam:
@karasusan Thank you! And thanks for the work on the feature, will make use of this right away

PoRain:
for the develop branch that you recently submitted, I can't run audiosample.cs normally. The specific performance is that I can't run audiosample.cs until the_ PC1 = new rtcpeerconnection (Ref configuration) ", unity crash. I guess there's a problem with the C # interface "contextcreatepeerconnectionwithconfig"， because when i uses the "_ pc1 = new rtcpeerconnection() ", unity will not break

nvanofwegen:
It's because the plugins/webrtc.dll file doesn't have the changes (yet), so the audio methods don't exist and new RTCConfiguration with optional fields are not compatible. For now I'd say the solution is to build the webrtc.dll yourself using the buildscripts.

PR #457, opened 2021-07-07
feat: Add audio renderer mute

rollersteaam:
The new audio send works like an absolute charm, besides the fact that microphone input must be heard by the user for it to be sent into a video call. This PR allows consumers to optionally not hear the microphone if they're sending it.


kannan-xiao4:
Hi, @rollersteaam .
Sorry late for reply. I check this pull-request.
Currently our implementation doesn't use `OnAudioFilterRead` callback.

So, if you want to mute the sending audio, turn on the Mute flag of `AudioSource` component used for the Sender Track.
Then you can mute the local audio and send the audio to remote.

karasusan:
HI @rollersteaam,
Could you have any response?

rollersteaam:
Heya! I can see the implementation for audio rendering has changed quite a bit. @kannan-xiao4 I haven't managed to test your suggestion but it sounds like it will work, so I'll close this :)

Cheers

Issue #470, opened 2021-07-14
[REQUEST] WebGL platform support


karasusan:
# Precondition

WebGL support has been advanced by external developers in Unity community.
Very thanks @nvanofwegen and @gtk2k
https://github.com/The-Barn-Games/com.unity.webrtc/tree/feature/webgl-support

WebGL support is a long journey and we have many issues with merging it to the main branch. So I made the feature branch that receiving contributions from developers.
https://github.com/Unity-Technologies/com.unity.webrtc/tree/experimental/webgl-platform

# TODO

- [ ] Create a new branch that developing for WebGL platform (@karasusan)
- [ ] Set up CI process of the WebGL platform in the internal CI platform ((@karasusan)
- [ ] Create a first PR to the feature branch (@nvanofwegen)
- [ ] Pull Request review (@karasusan, @kannan-xiao4)
- [ ] Manage GitHub issues (@karasusan)

karasusan:
This is the first PR
https://github.com/Unity-Technologies/com.unity.webrtc/pull/478

karasusan:
> Set up CI process of the WebGL platform in the internal CI platform ((@karasusan)

I tried this topic last week, however, I found this is needed more time than I expected.
So we test the package for WebGL platform manually.

karasusan:
memo: WRS-158

karasusan:
The pull request(#554) above was merged into this feature branch on our repository.
https://github.com/Unity-Technologies/com.unity.webrtc/tree/experimental/webgl-platform

Please feel free to try WebGL build using this branch. And post issues on this thread if you find.

karasusan:
The issue below is posted in Unity forum 
https://forum.unity.com/threads/webgl-platform-support.1147331/#post-7884949

Thaina:
Is it possible to add support for microphone in browser too?

karasusan:
@Thaina
I guess the `MediaStreamAddUserMedia` API is not supported for microphone.
Do you have any ideas? @nvanofwegen

Thaina:
As far as I can see `MediaStreamAddUserMedia` was called `navigator.mediaDevices.getUserMedia` which seem totally about microphone so I wish it did work and I am started to try using it now. But because it has no sample I still try to figure out how to start with it

[...]

PR #478, opened 2021-07-17
Experimental WebGL implementation

nvanofwegen:
First and foremost, thanking @gtk2k for making the fundamental code which was used to make this contribution.

## Summary
This PR adds experimental functionality for the WebGL platform, including:
- Data channels
- Receiving audio/video tracks
- Sending video with limited performance
- Sending audio not implemented yet
- RTCParameters and RTCStats are using complete JSON (de)serialization instead of pointers/marshalling, so the performance is low. For now, [jillejr.newtonsoft.json-for-unity](https://github.com/jilleJr/Newtonsoft.Json-for-Unity "jillejr.newtonsoft.json-for-unity") is required
- [getUserMedia](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia "getUserMedia") to send webcam/microphone over WebRTC (added to Samples)
- [adapter.js](https://github.com/webrtc/adapter "adapter.js") to improve browser support

A demo can be accessed here: https://d1ddjbgn3ezoae.cloudfront.net

### Audio
- Audio in WebGL is [limited](https://docs.unity3d.com/Manual/webgl-audio.html "limited"). Currently the AudioStreamTracks are not sending any data over WebRTC, as [OnAudioFilterRead](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnAudioFilterRead.html "OnAudioFilterRead") is not available

### Video
- Using [Camera](https://docs.unity3d.com/ScriptReference/Camera.html "Camera") to generate VideoStreamTracks has limited performance and a lot of WebGL warnings appear
- Receiving video has a colorspace issue (rendering slightly too white)
- (Samples) Using the framebuffer causes the UICamera to crash and the UI turns black
- When receiving multiple VideoStreamTracks OnTrack simultaneously events, some tracks can remain grey and are not initialized properly

### Firefox
- The [RTCStatsReport](https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport "RTCStatsReport") generated by Firefox contains less information
- [getCapabilities()](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender/getCapabilities "getCapabilities()") is not implemented in Firefox, so setting the codec is not available.
- Firefox has a loopback error when the hostname is `http://localhost:port` (which is the hostname when you Build and Run and Unity hosts the files). However, the hostname `http://localhost:port/path` works. So serving the files with your own toolkit can be used as a workaround (eg jetbrains/vscode/apache/nginx)

## Improvements/todo
- Find a way to generate and send audio sources. Alternatively, incoming audio could be implemented using the [Web Audio API](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletProcessor "Web Audio API") to send the audio data to Unity, or the DOM could be responsible for the audio and Unity can control the HTMLAudioElement
- Improve Camera video rendering performance
- Update RTCStats and RTCParameters to use pointers/marshalling
- Some RTCErrors are ignored
- Numerous TODO tags in the cs/jslib files
- The implementation is not fully tested, so expect unknown issues/bugs


unity-cla-assistant:
[CLA assistant check](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.webrtc?pullRequest=478) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.cds.internal.unity3d.com/Unity-Technologies/com.unity.webrtc?pullRequest=478) before we can accept your contribution.<br/>**0** out of **3** committers have signed the CLA.<br/><br/>:x: nvanofwegen<br/>:x: antoined73<br/>:x: Antoine Dezarnaud<br/><hr/>**Antoine Dezarnaud** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.cds.internal.unity3d.com/check/Unity-Technologies/com.unity.webrtc?pullRequest=478) it.</sub>

karasusan:
It seems This pull request is contributed by two developers.

@antoined73 
Could you sign the CLA?

antoined73:
I have added the mail corresponding to my other mail adress (Antoine Dezarnaud) and signed the CLA

karasusan:
@antoined73
Thanks, There are many things I don't know about the CLA system 😓 
But I checked you have already signed so it is no problem.

package.json
```diff
@@ -10,7 +10,8 @@
   ],
   "dependencies": {
     "com.unity.modules.jsonserialize": "1.0.0",
-    "com.unity.editorcoroutines": "1.0.0"
+    "com.unity.editorcoroutines": "1.0.0",
+    "com.unity.nuget.newtonsoft-json": "2.0.0"
```
karasusan, line 14: We have to remove the dependency for releasing the package as verified because the verified package must not depend on non-verified ones.

nvanofwegen: Do you mean the newtonsoft dependency has to be removed/resolved before the PR can be accepted? Because that requires the RTCParameter and RTCStats marshalling/ptr to be implemented, which is a big task.

Also, this line is wrong, as the code requires `jillejr.newtonsoft.json-for-unity": "13.0.102"` for now. The package `com.unity.nuget.newtonsoft-json": "2.0.0` crashed when parsing the RTCStats object due to IL2CPP/AOT issues.

This is why I added the `jillejr.newtonsoft.json-for-unity": "13.0.102"` as a limitation/restriction.

However, setting the package.json file requires the projects manifest.json file to be changed (scoped registry). So I'm not sure what to do now.

karasusan: Thanks, I understood.
I am investigating we can use the `com.unity.nuget.newtonsoft-json` package for our product.

karasusan: As you said, it is not clear this package supports IL2CPP or not.
https://docs.unity3d.com/Packages/com.unity.nuget.newtonsoft-json@2.0/manual/index.html

nvanofwegen: As far as I know, newtonsoft is currently used to following ways:

`com.unity.nuget.newtonsoft-json": "2.0.0"` is needed in the NativeMethods, for example:
```
#if !UNITY_WEBGL
        [DllImport(WebRTC.Lib)]
        public static extern void PeerConnectionCreateOffer(IntPtr ptr, ref RTCOfferAnswerOptions options);
        [DllImport(WebRTC.Lib)]
        public static extern void PeerConnectionCreateAnswer(IntPtr ptr, ref RTCOfferAnswerOptions options);
#else
        [DllImport(WebRTC.Lib)]
        public static extern void PeerConnectionCreateOffer(IntPtr ptr, string options);
        [DllImport(WebRTC.Lib)]
        public static extern void PeerConnectionCreateAnswer(IntPtr ptr, string options);
#endif
```

So we need the json serialization for these kind of methods.

The third party lib `jillejr.newtonsoft.json-for-unity": "13.0.102"` was only needed for the (de)serialization of RTCStats/Parameters because `com.unity.nuget.newtonsoft-json": "2.0.0"` crashed when deserializing the RTCStatsReport object (due to IL2CPP).

I think it's possible to remove the `jillejr.newtonsoft.json-for-unity": "13.0.102"` dependency eventually, but I am not sure if we can remove `com.unity.nuget.newtonsoft-json": "2.0.0"`.

Of course, there is JsonUtility, but yea... not sure it that is the right tool for this

gtk2k: 原文
JsonUtility cannot serialize / deserialize only arrays.
This is the only reason I used the jillejr package.
JsonUtility will give an error if it is an array, but arrays in an object can be serialized / deserialized.
If you set the rule that the data to be marshalled must be of object type, I think it's okay to use JsonUtility.
If you run into problems with this rule, why not consider the specifications of other packages at that point and consider using JsonUtility first?

```cs
private class Data
{
    public string[] alice;
    public float[] bob;
    public string candy;
    public int dad;
}

void Start()
{
    // Array Error.
    //var arr = JsonUtility.FromJson<Data[]>("[{\"alice\":[\"fuga\", \"piyo\"], \"bob\":[1.23, 3.45, 6.78], \"candy\":\"JsonUtility\", \"dad\": 123}]");
   
    // Object Success.
    var data = JsonUtility.FromJson<Data>("{\"alice\":[\"fuga\", \"piyo\"], \"bob\":[1.23, 3.45, 6.78], \"candy\":\"JsonUtility\", \"dad\": 123}");
    Debug.Log(data.alice);
    Debug.Log(data.bob);
    Debug.Log(data.candy);
    Debug.Log(data.dad);
}
```

nvanofwegen: If I understand it correctly, we theoretically can only use JsonUtility, with the rule that it is an Object, instead of Array. That's cool, I didn't know that.

But what do you mean with the last sentence? I think I'm confused by "consider other packages and consider JsonUtility first". I think it's fine to use other packages, and forcing everything to JsonUtility seems pretty difficult.

And just to be sure @karasusan, is there a problem for this PR that we are using a third party json lib for now? I understand that for the final version (merge to develop) it's required to use only unity packages, but removing the json lib dependency is a lot of work and would be better for another issue/PR.

gtk2k: What I wanted to say is that it's a bit annoying, but I think it can be implemented with the Json Utility. Consider using other JSON libraries in future versions. about it.

karasusan: @nvanofwegen

> I understand that for the final version (merge to develop) it's required to use only unity packages

Right, this is hard but we have to do it.

> but removing the json lib dependency is a lot of work and would be better for another issue/PR.

Sure, no problem. This is what we will do.

nvanofwegen:
Marked as draft again, merge commit is only resolving build errors. 
Build looks okay, but obviously audio is throwing a lot of errors (but that's out of the scope of this PR).

However, I still need to verify the following methods in webrtc context:
- Creating video/audio track source
- Add/Delete RefPtr methods

Runtime/Plugins/WebGL/VideoRenderer.jslib
```diff
@@ -0,0 +1,58 @@
...
+  },
+  
+  GetVideoRendererId: function (sinkPtr) {
+  },
+
+  DeleteVideoRenderer: function (contextPtr, sinkPtr) {
+
+  },
+
+  UpdateRendererTexture: function (trackPtr, renderTexturePtr, needFlip) {
```
karasusan, line 26: The `needFlip` argument is not used in this method.
I would like to remove this argument.

Runtime/Scripts/WebRTC.cs
```diff
@@ -804,36 +918,56 @@ internal static class NativeMethods
         public static extern void PeerConnectionRestartIce(IntPtr ptr);
         [DllImport(WebRTC.Lib)]
         public static extern RTCErrorType PeerConnectionSetConfiguration(IntPtr ptr, [MarshalAs(UnmanagedType.LPStr, SizeConst = 256)] string conf);
+#if !UNITY_WEBGL
         [DllImport(WebRTC.Lib)]
         public static extern IntPtr ContextCreateDataChannel(IntPtr ptr, IntPtr ptrPeer, [MarshalAs(UnmanagedType.LPStr, SizeConst = 256)] string label, ref RTCDataChannelInitInternal options);
+#else
+        [DllImport(WebRTC.Lib)]
+        public static extern IntPtr ContextCreateDataChannel(IntPtr ptr, IntPtr ptrPeer, [MarshalAs(UnmanagedType.LPStr, SizeConst = 256)] string label, [MarshalAs(UnmanagedType.LPStr, SizeConst = 256)] string optionsJson);
```
karasusan, line 926: I am not familiar the interoperability for WebGL and Unity, 😛 
That's why sorry to ask such a basic question.

- Can not we use `ref`, `out` keyword to communicate WebGL plugin ?
- I am sure that we need to use JSON to pass the struct data to WebGL plugin from Unity. Is there any other ways to pass them?

gtk2k: You can use ref and out, but since they are passed to JS with pointers, you can basically only pass values as numbers, strings, or arrays of numbers.
Also, the method of passing the structure can be done by creating a Serializer / Deserializer by yourself other than JSON, but the performance of the JSON API built in JS is better than the poor implementation of Serializer / Deserializer by yourself. 

(日本語)
一応、ref や out は使えますが、JSにはポインターで渡されるので基本的に数値か文字列、もしくは数値配列でしか値を渡せません。
構造体を渡す方法は、バイト配列を渡すことが可能なので、自分でSerializer/Deserializerを作成することでJSON以外でも可能にはなりますが、下手に自分でSerializer/Deserializerを実装するよりもJS組み込みのJSON APIのほうがパフォーマンスがいいです。

Runtime/Plugins/WebGL/VideoRenderer.jslib
nvanofwegen: Sure, will remove it! 

I kept it in place since I was fighting a bit with the default flipping of unity. In theory you could _not_ flip manually, so an y-inverted image is send over webrtc and have it (due to default flipping) correctly shown on the receiver.
But this becomes a problem when using a webcam mediastream, as they should be manually flipped on the receiver because that image is not y-inverted on the webrtc stream. So in order to send the correct image over webrtc, I always flip.

Runtime/Scripts/WebRTC.cs
nvanofwegen: As gtk2k said, we can only send basic types to jslib.  [Here](https://docs.unity3d.com/Manual/webgl-interactingwithbrowserscripting.html) it states the supported types (strings, different numbers in different HEAP arrays and a webgltextures map).
So unfortunately, sending json objects (strings) back and forth it the only way to go (at least to my knowledge) for these type of connectivity.

I did found on the emscripten docs that it's possible to send an array of pointers, which could be useful for the larger webrtc stats and/or rtc parameter classes and let the strings stay in the javascript heap and acces them from unity with pointers.

Runtime/Plugins/WebGL/VideoRenderer.jslib
karasusan: > But this becomes a problem when using a webcam mediastream, as they should be manually flipped on the receiver because that image is not y-inverted on the webrtc stream. So in order to send the correct image over webrtc, I always flip.

Certainly, This is the problem. Should we add the API to control whether flip the streaming video?

Runtime/Scripts/WebRTC.cs
karasusan: @gtk2k, @nvanofwegen
Talking to you is very helpful to me. Thanks.

I would like to reduce conditional branches using **UNITY_WEBGL** in `WebRTC.cs` and `Context.cs` to keep code easy to maintenance. Do you have any ideas to achieve it? There is no problem for me to rewrite trunk source code for it.

Runtime/Plugins/WebGL/VideoRenderer.jslib
nvanofwegen: I'm confused, do you want to be able to flip the image or not, both receiver/sender? 
At least on the receiving side I would always flip, as we should be expect the incoming stream to be correct and not flipped.
As a sender, it might be good to have the flip option I guess, as unity flips it but webcam stream doesn't. I'm not sure if there are other cases where unity webgl sends the non-flipped image.

Runtime/Scripts/WebRTC.cs
nvanofwegen: Yeah, the conditional branches are a pain... But I'm not sure if there is a good solution that is worth the struggle, just to have the code a little bit more maintainable. For now I'd say it is the best to have it better structured in WebRTC.cs, so the overview is a bit more clear, which makes it also easier the maintain. Right now the NativeMethods have the UNITY_WEBGL all over it, I agree this is a bit confusing.

What I found/think:
As unity and javascript have their own memory heap, it's afaik not possible to simply 'share' the objects between unity and javascript. Only the available heap arrays are an option (numbers, bytes, textures).  Strings are not shared as far as i know, just pointers to read the data.
gtk2k said that you could write a serializer/deserializer yourself, instead of json, using byte heap/arrays. Regarding that, also found [this thread](https://forum.unity.com/threads/source-code-c-javascript-interoperability-library.968487/) where a serializer/deserializer is used instead of json.

But, as this is not really my playing field, I have no idea if this requires a lot of c++ changes or not. It also looks like the functionality that is needed for the RTCParameter/RTCStats adoption, so maybe in the long run this has to be done anyway. If that is the case, I guess also implementing the ref interop functionality is pretty much the same?

Some other interesting thread I found, but it is maybe a bit outdated:
https://forum.unity.com/threads/trying-to-share-float-array-between-c-and-webgl-browser-javascript.394861/

[Here](https://forum.unity.com/posts/2579401) the threads starts about strings being immutable and [here](https://forum.unity.com/posts/2586907) it's even stated that it's simple not possible. But, as this thread is from 2016, so maybe now it's possible?

karasusan: Thanks.

> Here the thread starts about strings being immutable and here it's even stated that it's simple not possible. But, as this thread is from 2016, so maybe now it's possible?

I am not sure that the currently status but it would not be supported yet.

karasusan:
Sorry for the late response.
We will make a new branch for this pull request on the repository to test on Internal CI system.

karasusan:
I ran unit-tests on CI system and got an error below.
```
[07:47:16.370 Information] [5/6] TestStart 'Unity.WebRTC.EditorTest.PluginTest.CheckPluginImportSettings'

[07:47:16.467 Information] [5/6] TestFinish
	name: Unity.WebRTC.EditorTest.PluginTest.CheckPluginImportSettings
	result: FAILED
	message:   Expected: True
  But was:  False

	stackTrace: at Unity.WebRTC.EditorTest.PluginTest.CheckPluginImportSettings () [0x00054] in C:\udF09P\
[07:47:16.475 Information] Library\PackageCache\com.unity.webrtc@e3f6f4743af4-1631777616000\Tests\Editor\PluginTest.cs:45
```

There are two options to fix this issue.

- Enable the `Load on startup` flag for all files in the `Runtime/Plugins/WebGL` folder
image

- Fix the `PluginTest.CheckPluginImportSettings` test-case like below

```csharp
            if(!path.Contains("WebGL"))
            {
                Assert.IsTrue(pluginImporter.isPreloaded);
            }
```

Tests/Runtime/Unity.WebRTC.RuntimeTests.asmdef
```diff
@@ -1,13 +1,13 @@
...
     "name": "Unity.WebRTC.RuntimeTests",
+    "rootNamespace": "",
     "references": [
         "Unity.WebRTC",
         "UnityEngine.TestRunner",
         "UnityEditor.TestRunner"
     ],
     "includePlatforms": [
         "Android",
-        "Editor",
```
karasusan, line 10: This line is needed to run unit-test on Editor.

karasusan: Could you merge this commit into your branch?
https://github.com/Unity-Technologies/com.unity.webrtc/commit/187d9f324a93fd21c1688900b5782d74be59feef

karasusan: These errors occurred when running runtime tests.
image

Runtime/Scripts/WebRTC.cs
```diff
@@ -909,14 +1089,35 @@ internal static class NativeMethods
         [DllImport(WebRTC.Lib)]
         [return: MarshalAs(UnmanagedType.LPStr)]
         public static extern string IceCandidateGetSdpMid(IntPtr candidate);
+#else
+        [DllImport(WebRTC.Lib)]
+        public static extern string IceCandidateGetCandidate(IntPtr candidate);
+        [DllImport(WebRTC.Lib)]
+        public static extern int IceCandidateGetSdpLineIndex(IntPtr candidate);
```
karasusan, line 1096: Is this code needed?
It seems the same method above.

Also `IceCandidateGetSdp` and  `IceCandidateGetSdpMid` are same from original one.

nvanofwegen: This seems to be the case indeed, so the code is not needed and can be put outside of the UNITY_WEBGL guard. Only `NativeMethods.IceCandidateGetCandidate` is different.

karasusan: I understood. Thanks.

karasusan:
We reviewed this pull request and we concluded that it is necessary to fix the design for WebGL platform support.

The current implementation reduce maintainability. We should separate the platform-dependent implementation from the package source code.

As you know, WebGL platform is not supported below so we have to make other plugin interface.

- byte array
- pointer of the struct data
- multithread

How to fix:
- Because of the constraint of the plugin interface, we need to make the layer to close the gap between them
- Make that package developers don't concern the difference between platforms when using plugins

image (14)

If you agree this, I will make a new action plan.

nvanofwegen:
I already noticed some commits on feature/refactor-webgl-support, is that branch the implementation of your suggestion?
I never really tried to optimize the usability/maintainability of the code structure, I wanted to make it work/functional first.

I fully agree on the suggested code structure, the platform code should not be mixed with the package code. The structure you made on that branch actually made me smile :D.

If there is anything I can help with, please let me know. 

PS. Nice drawing!

karasusan:
> I already noticed some commits on feature/refactor-webgl-support, is that branch the implementation of your suggestion?
I wanted to make refactor but the work is so hard... I possibly won't finish by the end.

I am glad to hear that you agree my suggestion. I will consider the next action.

karasusan:
I tried to set up auto-testing for WebGL platform. However, not succeeded yet because the test runner doesn't work correctly as I expected. I am asking about the issue the WebGL support team.

karasusan:
As I posted the comment on the issue, I am testing the PR using Unity Test Framework.
https://github.com/Unity-Technologies/com.unity.webrtc/issues/470#issuecomment-927439719

These tests haven't passed yet, so we should fix them. To test the package for WebGL platform, run tests with the standalone suite. Please push the `Run All Tests` button like below:
image

nvanofwegen:
image
It's something! At least the test runner runs and gives results

Thing worth mentioning:
- [TestCase] attribute doesn't work, so I disabled those tests (not supported due to System.Reflection in NUnit)
- Excluded all [UnityTest] tests
- Disabled playing audio
- NativeAPI test class is still todo, disabled for now

karasusan:
@nvanofwegen 
Great work! 

> * [TestCase] attribute doesn't work, so I disabled those tests (not supported due to System.Reflection in NUnit)

Really? What kind of problem? I will also check the issue.

nvanofwegen:
When running the tests with [TestCase] I got the fatal error message:
> uncaught exception: abort("To use dlopen, you need to use Emscripten's linking support, see https://github.com/kripken/emscripten/wiki/Linking") 

When inspecting the stacktrace I got to the following lines which probably cause the error (as the checkIfIsMacOSX method was the last human readably method i could find in the trace).
https://github.com/nunit/nunit/blob/ca56293c2918d9ef9ace77f62a63842e59640b58/src/NUnitFramework/framework/Internal/OSPlatform.cs#L367

The pointer for dlopen() also was pointing to 'libc', so pretty sure this is the reason why it crashed if the custom unity nunit package also uses this script.

Question remains why [TestCase] triggers this script eventually.

nvanofwegen:
Woops, pushed with another git user. I added the other email to my github account for the CLA.

karasusan:
@nvanofwegen 
I reproduced the issue about the Unity Test Framework with WebGL platform. Thanks.
I already reported the issue into the QA team.

And I concerned how to avoid this issue and found **TestCaseSource** attribute that can do parameterized test like TestCase attribute without the issue. I am considering the change for testing on WebGL.

nvanofwegen:
Regarding the JSON error, is it possible that you are using the unity's json library, instead of `jillejr.newtonsoft.json-for-unity`?

CreateAudioTrack is defined in AudioStreamTrack.jslib, and is called by Context.jslib which seems to be included in your build... So I don't understand why AudioStreamTrack.jslib isn't included...

karasusan:
Installed `nuget.newtonsoft-json` in this package
image

You mean  `jillejr.newtonsoft.json-for-unity` should be install from below?
https://github.com/jilleJr/Newtonsoft.Json-for-Unity

karasusan:
@nvanofwegen
Surely **Strip Engine Code** flag was disabled, and I enabled this.
But the issue was not fixed. I am investigating for debug. Thanks.

nvanofwegen:
Looks good! I prefer the link.xml file, so all the information about what needs to be preserved is in one place/file, but this works. 

I also fixed some of the errors you faced, but there is an issue with the error in the SetDescriptionThrowException testcase:

When setting the description:
```  
PeerConnectionSetDescription: function (peerPtr, typeIdx, sdpPtr, side) {
    var peer = UWManaged[peerPtr];
    var type = UWRTCSdpType[typeIdx];
    var sdp = Pointer_stringify(sdpPtr);
    peer['set' + side + 'Description']({type: type, sdp: sdp})
        .then(function () {
      uwcom_debugLog('log', 'RTCPeerConnection.jslib', 'PeerConnectionSetDescription', peer.label + ':' + side + ':' + type + ':' /*+ sdp*/);
      Module.dynCall_vi(uwevt_OnSetSessionDescSuccess, peer.managePtr);
    }).catch(function (err) {
      uwcom_debugLog('error', 'RTCPeerConnection.jslib', 'PeerConnectionSetDescription', peer.label + ':' + side + ':' + err.message);
      var errorNo = uwcom_errorNo(err);
      var errMsgPtr = uwcom_strToPtr(err.message);
      Module.dynCall_viii(uwevt_OnSetSessionDescFailure, peer.managePtr, errorNo, errMsgPtr);
    });
    
    // TODO: Use promises to wait for resolve/reject and return RTCErrorType
    return uwcom_arrayToReturnPtr([UWRTCErrorType.indexOf("None"), uwcom_strToPtr("no error") ], Int32Array);
  },
```

Since the setDescription is asynchronous, jslib cannot wait to return the correct RTCErrorType. So the test case fails and disposes the RTCPeerConnection. But a brief moment later, the promise is resolved and tries to call the OnSetSessionDescFailure callback, but this callback is not there anymore because the testcase already ended.

However, I haven't found a way to use async/await inside jslib (Unity 2021 will allow this I think, as emscripten supports ES6).

I do get another error after the test suite is finished and Unity tries to quit:
image

karasusan:
👍 
image

karasusan:
I checked some samples and found bugs.

## Menu scene
Got error when clicking WebGL button.
```
Scene 'WebGL' couldn't be loaded because it has not been added to the build settings or the AssetBundle has not been loaded.
To add a scene to the build settings use the menu File->Build Settings...
```
image

## PeerConnection scene
Texts are broken.
image

## DataChannel scene
Got error when clicking `Back` button after clicking `Call` button.
```
[jslib] onclose: Unmanaged dataChannel. Ptr: 30
uwcom_existsCheck @ build.framework.js.gz:3424
(anonymous) @ build.framework.js.gz:3424
```

## Stats scene
Texts are broken.
image

## PerfectNegotiation sample
Got error when clicking `Swap` button.
```
RTCErrorException: Exception of type 'Unity.WebRTC.RTCErrorException' was thrown.
  at Unity.WebRTC.RTCPeerConnection.SetLocalDescription () [0x00000] in <00000000000000000000000000000000>:0 
```

nvanofwegen:
Unity 2020.3.1f1
Win10 Chrome M95

Menu scene
I guess my project setup is a bit different, as I copied the Samples folder into the project folder. But aren't build settings done on a project level? So how is this fixed?

PeerConnection
Also, strange, as my text is fine. However, you don't seem to have a bug which I have:
image
My UI background camera crashes/goes black, no idea why.

RTCStats
Same as PeerConnection, text is fine here...

DataChannel
I can reproduce this, will fix that asap :)

PerfectNegotiation
Also reproduced here, will check that out

nvanofwegen:
I fixed the issue in PerfectNegotiation, there was missing some implementation for SetLocalDescriptionWithoutDescription.

---

About the DataChannel error, I'm not sure if it's really an 'error'. It only happens when you call WebRTC.Dispose(), but it doesn't happen when you only close the PeerConnection and a few seconds later call WebRTC.Dispose().

I think the reason is that the onclose event on the datachannel is fired when calling RTCPeerConnection.Close(), but _not immediately_. When closing the peer connection, the onclose event triggers DataChannelNativeOnClose method, which removes the datachannel correctly on both pointer arrays.

But when calling WebRTC.Dispose(), we delete the peerconnection from the table before the onclose event is handled (which is triggered async). And it crashes at calling the DataChannelNativeOnClose method, because the datachannel is already removed from the pointer array (UWManaged).

So it only happens when you call WebRTC.Dispose(), which deletes the DataChannel from the pointer array before the onclose event is fired. So not having the datachannel in the pointer array is in this case not an error, because we tried to remove it twice.

In my opinion logging a warning is more fitting.

---

Maybe offtopic, but:
I found something that I find a bit strange in the base branch, where webgl and native code differ:
`ContextDeletePeerConnection` and `Context.DeletePeerConnection` are never referenced in C#, expect in some TestCases. However, I do call it in WebGL (RTCPeerConnection:89):
```
#if UNITY_WEBGL
                NativeMethods.ContextDeletePeerConnection(WebRTC.Context.self, GetSelfOrThrow());
#endif
```

Theoretically, I can remove this call, but then the actual PeerConnection object will remain in the managed array, which is also not a good thing?
So a bit offtopic in a WebGL PR, but for native builds, what happens with the PeerConnection objects after disposing them in C#? As the following native methods seem to be not called from the C# side:
https://github.com/Unity-Technologies/com.unity.webrtc/blob/2b34a3a0066ae35a9e93d53a9ac899f214a4c44b/Plugin~/WebRTCPlugin/WebRTCPlugin.cpp#L483 and https://github.com/Unity-Technologies/com.unity.webrtc/blob/2b34a3a0066ae35a9e93d53a9ac899f214a4c44b/Plugin~/WebRTCPlugin/Context.cpp#L507

karasusan:
@nvanofwegen

Menu Scene
You need to add the scene to the list of **Scenes in Build** and commit the `EditorBuildSettings.asset` 
image

nvanofwegen:
Just curious, how is your project folder configured to develop? Do you use only this repository (and use WebRTC~), or is this repository included in another project git repo (using git submodules for example).

I never saw such relative paths/files, but Unity couldn't open the project (for example, Samples contained '../../Samples~').
So I copied the the actual folders/files to the WebRTC~ folder and made sure the guid was kept the same and added the WebGL sample to the build settings in WebRTC~. But this can surely be done in a more friendly way.

Maybe related, currently I have the com.unity.webrtc folder (this repo) in the Assets/Packages folder of my project, not in Packages. Because when building from Packages, the new scripts are not included in the build...

So do you know if Unity does something extra when changing code in the Packages folder with the Package Manager? Maybe this has something to do with the issue you had with the code stripping, which was solved adding the Preserve attribute.

karasusan:
@nvanofwegen
You can open the Unty project under `WebRTC~` folder. 
The package use the symbolic link to include `Sample~` folder. Please check whether the symlink is enabled on your environment.

note:
Unity2021.2.0f1 has a crash bug which using symlink on windows.

karasusan:
@nvanofwegen
Hi, any questions?

nvanofwegen:
Not really, WebGL scene was added to the build settings.

About the text error I have no idea (but seems to be related to the overall video issues) and the data channel error is not really an error (explained in previous comments)

So, I was more or less waiting for the next step ^^'

karasusan:
@nvanofwegen 
Sure. Sorry for your waiting. 
I will test samples on Unity2020.3LTS and Unity2021.2. 

Thanks.

nvanofwegen:
I updated the code to support 2021.2, could you run it on the internal CI system?

karasusan:
@nvanofwegen
Sure. Could you merge this change?
https://github.com/Unity-Technologies/com.unity.webrtc/commit/356cfefbe37c1b9a947e34e51c46a6b83c63cf37

mukaschultze:
On `PeerConnectionSetDescriptionWithoutDescription` the variable `side` should not be used since it's not passed as an argument the same way it is for `PeerConnectionSetDescription`, this is throwing errors on the latest commit.

mukaschultze:
I just made two PRs (https://github.com/The-Barn-Games/com.unity.webrtc/pull/12#issue-1071529537 and https://github.com/The-Barn-Games/com.unity.webrtc/pull/13#issue-1071530861) fixing the issue above, and fixing another issue with binary message on data channels.

karasusan:
`NullPointerException` is thrown when opening Audio sample scene. As we know, audio support on WebGL is limited.
image

karasusan:
When pushing the `Back` button in DataChannel sample, we received the error below.
image

karasusan:
In Stats sample, we got error below.

1. Push `Call` button.
2. Push `Hangup` button.

image

nvanofwegen:
For audio: Not part of this PR, so errors are to be expected.

For data channels: as stated in [this comment](https://github.com/Unity-Technologies/com.unity.webrtc/pull/478#issuecomment-956813517) the 'error' is in this case, not an actual error.
Because the datachannel itself is already disposed, the datachannel that gets disconnected doesn't exist anymore in the reference table. It's basically a false error, which occurs by only destroying the context without cleanup first. However, I don't want to remove the comment, as the datachannel can be disconnected from the other side of the peer connection. It's simply prevented by hanging up first, then destroy the context.
Also, this scenario only occurs when you are connected to yourself, which is not a very likely scenario (but a valid scenario nonetheless) 

For stats: I'll take a look, hangup should clean properly, so that needs to be fixed.

mukaschultze:
There's a small inconsistency for data channels between the web and standalone. On the web the `OnOpen` callback is called for channels received via `RTCPeerConnection.OnDataChannel`, but on standalone it's not.

I suspect the web behaviour is the correct one, but I'm not sure.

karasusan:
@mukaschultze
Yes, web browser behavior is correct, so It might be the bug.
I will be so glad if you make the issue page and write more detail of that.

kannan-xiao4:
In `MultiPeers` sample, 
On push `Start` button, console logging below error.
```
OverflowException: Arithmetic operation resulted in an overflow
```
after push `Call` button, this log continue to be output.

And push `HangUp` ,  error pop below
```
Uncaught RuntimeError: null function or function signature mismatch
    at build.wasm.gz:0x62735d
    at build.wasm.gz:0x165dea2
    at build.wasm.gz:0x147a5ab
    at build.wasm.gz:0xe6c456
    at invoke_vii (build.framework.js.gz:3425)
...
    at invoke_viiii (build.framework.js.gz:3425)
...
    at invoke_iiii (build.framework.js.gz:3425)
...
    at browserIterationFunc (build.framework.js.gz:3425)
    at callUserCallback (build.framework.js.gz:3425)
    at Object.runIter (build.framework.js.gz:3425)
    at Browser_mainLoop_runner (build.framework.js.gz:3425)
```

enviroment
- GoogleChrome on Windows10
- Build by Unity 2021.2

kannan-xiao4:
In `Bandwidth` sample
On change parameter using dropdown menu, error happened.
image

but this error only occurred in app that build by Unity 2020.

kannan-xiao4:
In `PerfectNegotiation` sample
The following error has occurred in the second (and from then) Swaps.
```
[jslib] SenderReplaceTrack: Unmanaged track. Ptr: 0
```
Sample works fine, but I think `BAD_ACCESS` happen.

karasusan:
@mukaschultze
Hi, how is the situation here?

nvanofwegen:
Just got back from a long vacation, so I didn't have the time to look at it yet.

As for the stats issue and the issues @kannan-xiao4 mentioned, I will inspect those probably somewhere this week.

However, I have some things about the webgl implementation in general.

This PR is open for half a year now, and every couple of weeks things were added that had to be fixed. Meanwhile, the last rebase is also 5 months old, so I'm pretty sure there are again a lot of issues before this PR can be merged.
As this is done in my own free time, it's starting to feel a bit annoying that small issues are added every now and then and the PR stays open without a real plan and it feels more like a chore now. 

So what is the roadmap for WebGL? For both this PR and the steps afterwards until it is ready to be merged in develop. The roadmap table in readme.md also hasn't been updated in a while.

I'm happy to look at the current issues found, but also patching the last 5 months of development where I'm the only one that can fix it is not something I'm looking forward to...

What is the downside of accepting the PR in the coming period, managing the issues here on github, work on the implementation and also let others be able to help?

karasusan:
@nvanofwegen
I apologize for taking your free time. Our team are discussing about your reply. 

FYI: I posted this thread for describing our direction.
https://forum.unity.com/threads/unity-render-streaming-status-update-q4-2021.1215801/

@gtk2k
@antoined73
@mukaschultze
If you have an opinion, I would like to hear it.

nvanofwegen:
No need to apologize! I picked up the work from gtk2k myself! I actually had a lot of fun developing so far. I only think that my lack of a clear understanding of how to proceed took the fun out a bit.

About the forum thread, awesome, haven't seen it and already clarifies a bit!

With WebGL being a community-based development for the foreseeable future, I do think the [proposed](https://github.com/Unity-Technologies/com.unity.webrtc/issues/470) approach should be in place rather sooner than later, as this PR is far from a 'release'.

In the thread you mention that people can see the progress in this PR, but this PR has become a bit of a dragon (80+ replies). There are several decisions in this PR made about audio, marshalling, refactoring the api, etc, but there is no overview of that. 

With those new managed issues (like refactoring, marshalling, audio, etc) discussions about the implementation could be better structured and organized and we could have a good overview in one main WebGL issue.

mukaschultze:
I've been using mostly the `DataChannels` to implement a transport for https://github.com/vis2k/Mirror. From my tests it works as well as the native implementation.

> With WebGL being a community-based development for the foreseeable future, I do think the [proposed](https://github.com/Unity-Technologies/com.unity.webrtc/issues/470) approach should be in place rather sooner than later, as this PR is far from a 'release'.

Totally agree with that

gtk2k:
Thank you for your contribution to WebGL platform support over the long term to this day.

About the three points @nvanofwegen mentioned, audio, marshalling, and refactoring

### audio
I think we are considering whether to implement this in the current PR. Since Microphone is not officially supported by Unity, it is possible to implement it by means such as using external assets, but I would like to avoid it as much as possible.

### marshalling
Earlier, @karasusan drew a diagram and commented
https://github.com/Unity-Technologies/com.unity.webrtc/pull/478#issuecomment-924563953
I recognize that this is the main thing.
Currently, I've put together the Plugin API calls in WebRTC.cs (using a preprocessor), but I think it's easier to implement by creating a separate file for WebGL (eg WebRTC_WebGL.cs) and implementing a layer for WebGL there. increase.

### refactoring
Refactor once all the tests pass with the current PR. After that, I think it would be better to consider audio and marshalling, but how about it?

If there is something I can help, I would like to help.

@karasusan
(日本語で失礼します)
私は、このリポジトリのフォルダ構成でUnityプロジェクトを開く方法がわかりません。WebRTC\~フォルダにAssetsフォルダなどのUnityプロジェクトフォルダ構成となっており、このWebRTC~フォルダをUnityで開いてみたのですが、以下の画像のエラーが発生します。
com unity webrtc_project_load_error

package.jsonには相対アドレスが書かれていますが、これは本来Unity側で認識してくれて、実際のファイルを読み込んでくれるものなのでしょうか？

ちなみに Windows (11)です。

試しに、Windowsでのシンボリックリンクに置き換えてみたのですが(Runtime, Editor, TestsフォルダもWindowsシンボリックリンクに変更)、今度はプロジェクトを開くと以下の状態でずっとローディングのままとなってしまいます。(CPU使用率が上がりっぱなしで、一応何らかの処理は行われているようですが。。。)
com unity webrtc_project_never_ending_load

どのようにすればプロジェクトを(Windows)Unityで開けるのでしょうか？

karasusan:
Sorry for everyone but writing in Japanese is easier for me 🙂 

@gtk2k
`WebRTC~` 以下を Unity プロジェクトとして開くことができます。
git のチェックアウト時にシンボリックリンクが正しく反映されなかった可能性があります。以下の記事を確認してください。
https://isgs-lab.com/440/

指摘のあった `package.json` は[これ](https://github.com/Unity-Technologies/com.unity.webrtc/blob/61d956800b28ecf157b6a99a0014d3b412032c6f/WebRTC~/Packages/com.unity.webrtc/package.json)のことだと思いますが、相対パスのように表示されているということは、git がシンボリックリンクをファイルとして扱っているということだと思います。

Windows11 が原因ではないとは思いますが、テスト環境が手元にないので確証はないです。

gtk2k:
@karasusan 
ありがとうございます。
Windowsの場合、git の config が複数存在し、また、アカウントを切り替えて使用しているため、Symlinks が false となっているのが原因で、これをtrueにすることでプロジェクトを開くことができました。

ですが、WebGLにプラットフォームを切り替えるときに Sampleの*.wavファイルで Import が進まなくなってしまいます。
現時点において、オーディオは実装されていないため、とりあえず*.wavファイルを削除することで回避してます。

karasusan:
@nvanofwegen
@gtk2k
@antoined73
@mukaschultze

Hi, our team discussed about this topic.

- We are going to merge this pull request to the experimental feature branch `experimental/webgl-platform` in our repository soon.
- After merged that, please keep developing based on the feature branch.
- About known issues for WebGL support, we expect that you create issues on GitHub Issues, and add the link for the [parent issue](https://github.com/Unity-Technologies/com.unity.webrtc/issues/470).

### What our team do

- Maintain CI system (building and testing)
- Review pull requests from external developers
- Check issue tickets

If you have questions, please feel free to ask me.
I appreciate your efforts.

PR #487, opened 2021-07-21
fix: Add UI to select DSP buffer size option in Audio sample

karasusan:
The DSP buffer size setting affects the audio latency, so I added the UI to select options.
image


kannan-xiao4 (approved):

PoRain:
Excuse me, I tried this sample with DSP. In the best delay mode, its performance is still not as good as the native webrtc. To be exact, its delay perception is obvious. I don't know if it's the problem of my device?

PR #496, opened 2021-07-30
fix: crash disposing track of sender

kannan-xiao4:


kannan-xiao4:
fixed by https://github.com/Unity-Technologies/com.unity.webrtc/pull/500

PR #500, opened 2021-08-04
fix: Manage lifespan of native objects using reference counter 

karasusan:
This pull request is for generalizing the PR below.
https://github.com/Unity-Technologies/com.unity.webrtc/pull/496

- To dispose of the native object safely, manage the lifespan of the native object from C# code
- Use `rtc::RefCountedObject` class to manage timing of disposing objects 
- Define **RefCountedObject** class in C# code which has the native pointer relating the `rtc::RefCountedObject` and decrement the reference count when the object is disposed of.

### DIFF
--- a/Plugin~/WebRTCPlugin/WebRTCPlugin.cpp
+++ b/Plugin~/WebRTCPlugin/WebRTCPlugin.cpp
@@ -280,15 +280,19 @@ extern "C"
     {
         context->StopMediaStreamTrack(track);
     }
-
-    UNITY_INTERFACE_EXPORT webrtc::MediaStreamTrackInterface* ContextCreateAudioTrack(Context* context, const char* label)
+    UNITY_INTERFACE_EXPORT webrtc::AudioSourceInterface* ContextCreateAudioTrackSource(Context* context)
     {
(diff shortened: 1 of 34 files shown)


Plugin~/WebRTCPlugin/WebRTCPlugin.cpp
```diff
@@ -276,24 +271,41 @@ extern "C"
...
+        Context* context, const char* label, webrtc::AudioSourceInterface* source)
+    {
+        return context->CreateAudioTrack(label, source);
     }
 
-    UNITY_INTERFACE_EXPORT::webrtc::MediaStreamTrackInterface* ContextCreateAudioTrack(Context* context, const char* label)
+    UNITY_INTERFACE_EXPORT void ContextAddRefPtr(Context* context, rtc::RefCountInterface* ptr)
     {
-        return context->CreateAudioTrack(label);
+        context->AddRefPtr(ptr);
```
kannan-xiao4, line 303: Is it no need to check about already registered in the reference map?

kannan-xiao4 (approved):

PR #503, opened 2021-08-05
fix: Audio channel count 

karasusan:
Sometimes audio streaming is received audio as a monoral channel even if sending the stereo.
This reason that the first few frames of audio are received as a monoral from the webrtc library.

This fix is ignoring the first few frames to get the right number of channels when receiving the audio streaming data.

### DIFF
--- a/Plugin~/WebRTCPlugin/WebRTCPlugin.cpp
+++ b/Plugin~/WebRTCPlugin/WebRTCPlugin.cpp
@@ -488,9 +488,15 @@ extern "C"
         obj->connection->RestartIce();
     } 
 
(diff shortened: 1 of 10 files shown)


pavelsamoylenko:
@karasusan hello!
When can the 2.4.0-exp.4 release with audio renderer be expected?

karasusan:
> @karasusan hello!
> When can the 2.4.0-exp.4 release with audio renderer be expected?

Sorry for the late reply.
This should be released this week. Thanks.

kannan-xiao4 (approved):

Issue #507, opened 2021-08-14
[REQUEST] Receive incoming video streams at their native resolution


nathanael-omnivor:
Currently, it appears that the only way to access the incoming video streams is by calling `InitializeReceiver(int width, int height)` on the `VideoStreamTrack`, providing your own `width` and `height`.

It would be great if we could access the incoming video at its native resolution, rather than provide a resolution manually. Or, at least provide a way to query for the video resolution, and then construct the receiver `Texture` using the existing method.

One complication is that the sending side's WebRTC implementation may dynamically change the resolution of the video it is sending, in which case the video received by Unity.WebRTC will change size as well. In this case, the receiver `Texture` would need to be updated to reflect the new resolution.

karasusan:
Thanks for your feedback. 
We will check your suggestion.

kannan-xiao4:
memo: WRS-139

karasusan:
@nathanael-omnivor
This issue is fixed by the next release which will release next week.

karasusan:
@nathanael-omnivor
We released the new version 2.4.0-exp.5.
Please check it out and if your issue is not solved, please notify me.

PR #517, opened 2021-08-25
feat: Implement video frame resize callback

kannan-xiao4:
Issue https://github.com/Unity-Technologies/com.unity.webrtc/issues/507

### DIFF
--- a/Runtime/Scripts/VideoStreamTrack.cs
+++ b/Runtime/Scripts/VideoStreamTrack.cs
@@ -82,6 +82,11 @@ namespace Unity.WebRTC
 
         internal void UpdateReceiveTexture()
         {
+            if (m_sourceTexture == null || m_destTexture == null)
+            {
+                return;
+            }
+
             // [Note-kazuki: 2020-03-09] Flip vertically RenderTexture
             // note: streamed video is flipped vertical if no action was taken:
             //  - duplicate RenderTexture from its source texture
--- a/Runtime/Scripts/VideoStreamTrack.cs
+++ b/Runtime/Scripts/VideoStreamTrack.cs
@@ -197,27 +197,15 @@ namespace Unity.WebRTC
 
         internal void OnVideoFrameResize(int width, int height)
         {
-            if (m_sourceTexture == null)
+            if (m_sourceTexture != null && (m_sourceTexture.width == width && m_sourceTexture.height == height))
             {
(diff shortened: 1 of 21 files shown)


doctorpangloss:
Does it make sense to allocate the render texture on the plugin side? It would work cleanly even if you have to `Release` and `Create` because the Unity runtime reference to the render texture would remain the same.

Runtime/Scripts/VideoStreamTrack.cs
```diff
@@ -184,6 +194,19 @@ public override void Dispose()
             }
             base.Dispose();
         }
+
+        internal void OnVideoFrameResize(int width, int height)
```
kannan-xiao4, line 198: @doctorpangloss 
Is your concern that this method indicate that you are creating a RenderTexture?

doctorpangloss: yes, this is going to leak textures if the width and height change.

kannan-xiao4: thanks
I understand and fixed in this commit ad70ecc

Runtime/Scripts/VideoStreamTrack.cs
```diff
@@ -184,6 +194,40 @@ public override void Dispose()
...
+            if (m_sourceTexture != null && (m_sourceTexture.width == width && m_sourceTexture.height == height))
+            {
+                return;
+            }
+
+            if (m_sourceTexture != null && m_sourceTexture is RenderTexture source)
+            {
+                source.Release();
+                source.width = width;
+                source.height = width;
```
karasusan, line 209:
```suggestion
                source.height = height;
```

Runtime/Scripts/VideoStreamTrack.cs
```diff
@@ -184,6 +194,40 @@ public override void Dispose()
...
+            else
+            {
+                m_sourceTexture = CreateRenderTexture(width, height);
+            }
+
+            if (m_destTexture != null)
+            {
+                m_destTexture.Release();
+                m_destTexture.width = width;
+                m_destTexture.height = width;
```
karasusan, line 221:
```suggestion
                m_destTexture.height = height;
```

karasusan (approved):

Issue #525, opened 2021-09-10
[BUG] high latency for inbound audio track


AndrewMeadows:
We're seeing high latency (up to 2.0 seconds) between speaking into the microphone of the remote peer and playing the audio at the receiving peer .  We think it is caused by a growing webrtc jitter buffer for the inbound audio track, however we are still trying to investigate and learn more.

**To Reproduce**
**(1)** Unity Peer1 on machine1 connect to C++ webrtc server
**(2)** Chrome browswer Peer2 on machine2 connect to C++ webrtc server
**(3)** Note: server does spatialization mixing and relays results to Peer1 and Peer2
**(4)** Peer1 speaks into mic...
**(5)** Peer2 hears audio with low latency (0.25 - 0.5 seconds)
**(6)** Peer2 speaks into mic ...
**(7)** ... Peer1 hears audio with high latency (1.0 - 2.0 seconds) **BUG!**

It is intermittent!  Sometimes we can get a lower latency (0.25 - 0.5 seconds) by doing the following:
**(8)** Mute the mic on peer2 and wait several seconds.
**(9)** Unmute the mic on Peer2 and speak...
**(10)** ... Peer1 sometimes hears audio with lower latency but eventually the latency grows to high levels again.

Note: we also tested on Windows but there only noticed high latency: we couldn't use the mute/unmute trick to get intermittent low latency audio.  We didn't try this for very hard: two or three times.  We mostly just wanted to verify that it wasn't a linux-only problem.

**Expected behavior**
The latency for audio from Peer1 to Peer2 is low (0.25 - 0.5 seconds).  We would expect similar latency in the other direction: Peer2 to Peer1.

**Theory: is Unity draining webrtc's audio buffer too aggresively?**
The behavior suggests the unity plugin is sometimes asking for audio from webrtc's inbound audio buffer too often, or in a bursty fashion, which could cause webrtc to supply "fake" audio data and grow its jitter buffer to compensate.  When inbound audio is muted webrtc has an opportunity to flush its jitter buffer to be smaller which is maybe why it sometimes shows up as low latency for a while.

We are trying to investigate.  The plan is to harvest webrtc stats to make a histogram of jitter buffer state.  If we learn more we will post updates.

**Environment:**
 - OS: Ubunutu-20.04
 - Browser: Chrome 92.0.4515.159 (for Peer2)
 - Unity Version: 2020.3.17f1 
 - Package version: 2.4.0-exp.4

[...]

PR #565, opened 2021-12-07
Reduce audio latency using buffer position from attached AudioSource

leejeonghun:
Hello.

As commented in #525, current audio playback has very high latency.
In order to resolve this, I tried to call `SetData()` of `AudioClip` with an offset which refers the sample position from attached `AudioSource`. It's a tricky way but works well.

Thank you in advance for your review.

### DIFF
--- a/Runtime/Scripts/AudioStreamTrack.cs
+++ b/Runtime/Scripts/AudioStreamTrack.cs
@@ -140,7 +140,7 @@ namespace Unity.WebRTC
                 while (numOfFrames-- > 0)
                 {
                     writtenSamples += WriteBuffer(
-                        m_recvBufs.Count > 0 ? m_recvBufs.Dequeue() : null,
+                        m_recvBufs.Count > 0 ? m_recvBufs.Dequeue() : new float[m_bufInfo.SamplesPer10ms * m_clip.channels],
                         baseOffset + writtenSamples);
                 }
 
@@ -148,7 +148,6 @@ namespace Unity.WebRTC
 
                 int WriteBuffer(float[] data, int offset)
                 {
-                    data ??= new float[m_bufInfo.SamplesPer10ms * m_clip.channels];
                     m_clip.SetData(data, offset % m_clip.samples);
                     return data.Length / m_clip.channels;
                 }
--- a/Runtime/Scripts/AudioStreamTrack.cs
+++ b/Runtime/Scripts/AudioStreamTrack.cs
@@ -1,5 +1,6 @@
 using System;
-using System.Collections.Concurrent;
+using System.IO;
+using System.Collections.Generic;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
 using UnityEngine;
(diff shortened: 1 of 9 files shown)


karasusan:
Wow, I will review it! Thanks a lot!

karasusan:
@leejeonghun 
Hi, I fixed the build error here and updating the plugin. It will be merged to develop branch soon.
Please review it and rebase this branch.
https://github.com/Unity-Technologies/com.unity.webrtc/pull/566

leejeonghun:
@karasusan 
#566 looks good to me. I'll rebase as soon as it merge.

leejeonghun:
@karasusan 
I rebased, please check again.

karasusan:
Pushed the plugins into this branch.
https://github.com/Unity-Technologies/com.unity.webrtc/tree/feature/low_latency

karasusan:
On Unity2019.3.33f1, `??=` operator is not allowed to use.

```
Packages\com.unity.webrtc\Runtime\Scripts\AudioStreamTrack.cs(151,28): error CS1525: Invalid expression term '='
```

Runtime/Scripts/AudioStreamTrack.cs
```diff
@@ -35,13 +36,63 @@ public AudioClip Renderer
...
         internal class AudioStreamRenderer : IDisposable
         {
-            private const int SamplesPerSec = 100;
-
             private AudioClip m_clip;
-            private readonly BlockingCollection<float[]> m_recvBufs = new BlockingCollection<float[]>(SamplesPerSec);
+            private bool m_bufferReady = false;
+            private readonly Queue<float[]> m_recvBufs = new Queue<float[]>();
+            private readonly AudioBufferTracker m_bufInfo;
+            private AudioSource m_attachedSource;
```
karasusan, line 95: `m_attachedSource` instance should be set from `AudioStreamRenderer` constructor.
There is a problem that finding the instance in `SetData` method.

Runtime/Scripts/AudioStreamTrack.cs
```diff
@@ -66,32 +117,72 @@ public void Dispose()
...
+            internal void SetData(float[] data)
+            {
+                m_recvBufs.Enqueue(data);
 
-                int remain = data.Length;
-                while (remain > 0)
+                if (m_recvBufs.Count >= AudioBufferTracker.NumOfFramesForBuffering && m_bufferReady == false)
                 {
-                    if (!m_recvBufs.TryTake(out float[] src))
+                    var audioSource = FindAttachedAudioSource();
```
karasusan, line 163: `m_attachedSource` should be set in the constructor.

Runtime/Scripts/AudioStreamTrack.cs
```diff
@@ -66,32 +117,72 @@ public void Dispose()
...
                 }
+            }
+
+            internal void SetData(float[] data)
+            {
+                m_recvBufs.Enqueue(data);
 
-                int remain = data.Length;
-                while (remain > 0)
+                if (m_recvBufs.Count >= AudioBufferTracker.NumOfFramesForBuffering && m_bufferReady == false)
```
karasusan, line 161:
```suggestion
                if (m_recvBufs.Count >= AudioBufferTracker.NumOfFramesForBuffering && !m_bufferReady)
```

Runtime/Scripts/AudioStreamTrack.cs
leejeonghun: > There is a problem that finding the instance in `SetData` method.
I'd like to ask what the problem is? because it's a intentional lazy initialization to wait until that m_clip attached by AudioSource.

leejeonghun:
> However, on Unity2019.3.33f1, `??=` operator is not allowed to use. Could you fix this error?

Sure, it's fixed from https://github.com/Unity-Technologies/com.unity.webrtc/pull/565/commits/3a13e626959518d373705c286225a8cd2c8e8d7e

Runtime/Scripts/AudioStreamTrack.cs
leejeonghun: It's fixed from https://github.com/Unity-Technologies/com.unity.webrtc/pull/565/commits/304c8508629f086cd72d70c4adba064cebb2fc3e

Runtime/Scripts/AudioStreamTrack.cs
karasusan: `GameObject.FindObjectsOfType` method is used in this line. 
https://github.com/Unity-Technologies/com.unity.webrtc/pull/565/files#diff-3eff72b49e750b2ef3b0f9311a838c24f98ddcfbd5482d1db99df5ce84d7a521R125

I feel setting `AudioSource` from constractor is better than using the Find* method.

leejeonghun: I'd like to ask you why is it better to use Find* method in constructor? I don't have much experience with Unity, so if you explain in more detail, I can understand your concerns.

karasusan: OK, please let me explain more detail.
My suggestion is making constructor having `AudioSource` argument, and removing `FindAttachedAudioSource` method.

```
public AudioStreamRenderer(AudioSource source)
```

Generally, `GameObject.Find*` has worse performance comparing to direct referencing. There are some projects restricting using  `GameObject.Find*` to avoid performance issue. Additionally, using  `GameObject.Find*` decrease readability and testability.

These links are just a few examples.
[https://dev.to/andytq/unity-comparison-of-performance-cost-between-gameobject-find-and-direct-referencing-pb](https://dev.to/andytq/unity-comparison-of-performance-cost-between-gameobject-find-and-direct-referencing-pb)
[https://akbiggs.silvrback.com/please-stop-using-gameobject-find](https://akbiggs.silvrback.com/please-stop-using-gameobject-find)
[https://gamedevunboxed.com/why-using-unitys-find-functions-is-a-mistake/](https://gamedevunboxed.com/why-using-unitys-find-functions-is-a-mistake/)

If you are OK, Could you allow me to refactor your code?

leejeonghun: As you said, to avoid using `FindObjectsOfType`, passing AudioSource as argument was considered too. but it makes breaking changes to it's usage, so I had to use `FindObjectsOfType` for current interface compatibility.
Anyway, please feel free to make any modification whatever it is, welcome better solution always. 😃

karasusan:
@leejeonghun
I created the branch for refactoring this pull request. Could you review it?
https://github.com/leejeonghun/com.unity.webrtc/pull/1

And we need to do rebase this change into the develop branch.

leejeonghun:
> And we need to do rebase this change into the develop branch.

@karasusan it is now rebased.

karasusan (approved):

kannan-xiao4 (approved):

Issue #559, opened 2021-11-28
[BUG] The Apple Silicon support is rarely showing picture when running in editor: never on Apple Silicon and extremely poorly on Intel Rosetta 2


doctorpangloss:
**Describe the bug**
Prior to recompiling `webrtc.bundle`, I could not get Apple Silicon to work in either Standalone Intel Rosetta 2, Standalone Apple Silicon, Editor Intel Rosetta 2, or Editor Apple Silicon. For some reason `MetalGraphicsDevice::ConvertFormat()` was returning `MTLPixelFormatInvalid`, according to the editor stack trace.

Recompiling fixed this issue, which was extra mysterious. I use a different `cmake` setup than you:

```
cmake . -G Xcode -D "CMAKE_OSX_ARCHITECTURES=arm64;x86_64" -B cmake-build-release -T buildsystem=1 -DCMAKE_CXX_FLAGS="-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk"
```

On Apple Silicon Editor, you will successfully create a peer connection and such, but picture will not appear in the browser, and the editor hangs.

On Intel Rosetta Editor, running the plugin in Release with the debugger attached showed a single still frame or two after a while in the browser. Then it crashes with an invalid access exception inside Apple libraries.

I successfully see a picture in a Standalone build, both on Intel Rosetta 2 and Apple Silicon.

**To Reproduce**
1. Run the Intel Rosetta 2 2021.2.4f1 and Apple Silicon 2021.2.4f1 editors and observe you do not see picture in the browser.
2. Build to standalone and observe you do see an H264 compressed stream (that eventually crashes but that's a different issue).

**Expected behavior**
Editor should correctly stream.

**Environment (please complete the following information):**
 - OS: macOS 11.15 Apple Silicon
 - Browser: Safari and Chromium
 - Unity Version: 2021.2.4f1
 - Package version: 2.4.0-exp4

karasusan:
Thanks for your sharing.
We know the crash occurred when using the package on Unity Editor Apple Silicon.

karasusan:
memo: WRS-153

doctorpangloss:
Tested and resolved by #568

PR #568, opened 2021-12-16
fix: Fix crash bug on Unity Editor 2021.2 on macOS(Apple Silicon)

karasusan:
Related issue.
https://github.com/Unity-Technologies/com.unity.webrtc/issues/559

Change the dynamic library format to `dylib` from `bundle` to fix the crash bug.

The cause of this crash is `UnityPluginLoad` function in the native code is not called.
This problem can be fixed if setting filename `webrtc` as a parameter of `DllImport` attribute instead of file path. When using `bundle` file, `DllImport` was failed and throw `DllNotFoundException`.


kannan-xiao4 (approved):

kannan-xiao4 (commented):
I think `macOS/libwebrtc.dylib` files not exists.
Do you forget to commit any files?

PR #575, opened 2021-12-24
feat: Upgrade NVIDIA Codec SDK 11.0

karasusan:


aet:
At some point I was looking into updating the SDK into the very latest, but I recall there was some driver issue holding me back when running on AWS. I couldn't upgrade further than 10.0.26, maybe things have changed since then?

karasusan:
Nothing particularly.
May I ask you about the issue more detail?

aet:
Tesla T4 driver version drags behind consumer hardware/desktop OS driver versions.

I think I was just experiencing a bad timing 12-18 months ago where using the latest SDK was requiring a more recent driver release than what was available for the cloud build, it's probably not an issue anymore but good to keep in mind while testing.

karasusan:
I understood that we should test it on Tesla T4 for upgrading the SDK. It is too troublesome...

kannan-xiao4:
Maybe NvCodec/NvDecoder does not update version same as NvEncoder.
https://github.com/Unity-Technologies/com.unity.webrtc/tree/feat/upgrade-nvcodec-11.0/Plugin~/NvCodec/NvCodec/NvDecoder

kannan-xiao4 (approved):

PR #591, opened 2022-01-13
fix: Flip received video frame vertically in native code

karasusan:
This change fixes the delay of video streaming receiver. The delay was 1 frame for flipping received texture vertically. 
This fix uses libyuv feature described [here](https://chromium.googlesource.com/libyuv/libyuv/+/HEAD/docs/rotation.md#invert-vertical-flip).


kannan-xiao4 (approved):
lgtm

PR #605, opened 2022-01-21
refactor: Refactor AudioStreamReceiver class that the class has a own native pointer

karasusan:
This pull request is for refactoring `AudioStreamReceiver`, which is `AudioStreamTrack` internal class.
This change makes clear the ownership about the audio rendering process in `AudioStreamTrack` class.

Please note that this change is not solved known issues, but only refactoring.

### DIFF
--- a/Plugin~/WebRTCPlugin/Context.h
+++ b/Plugin~/WebRTCPlugin/Context.h
@@ -66,11 +66,13 @@ namespace webrtc
         CodecInitializationResult GetInitializationResult(webrtc::MediaStreamTrackInterface* track);
 
         template <typename T>
-        bool ExistsRefPtr(T* ptr) {  return m_mapRefPtr.find(ptr) != m_mapRefPtr.end(); }
+        bool ExistsRefPtr(rtc::RefCountInterface* ptr) {
+            return m_mapRefPtr.find(ptr) != m_mapRefPtr.end(); }
         template <typename T>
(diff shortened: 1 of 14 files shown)


Plugin~/WebRTCPlugin/Context.h
```diff
@@ -168,7 +170,7 @@ namespace webrtc
         std::map<const webrtc::MediaStreamTrackInterface*, std::unique_ptr<VideoEncoderParameter>> m_mapVideoEncoderParameter;
         std::map<const DataChannelInterface*, std::unique_ptr<DataChannelObject>> m_mapDataChannels;
         std::map<const uint32_t, std::shared_ptr<UnityVideoRenderer>> m_mapVideoRenderer;
-        std::map<webrtc::AudioTrackInterface*, std::unique_ptr<AudioTrackSinkAdapter>> m_mapAudioTrackAndSink;
+        std::map<AudioTrackSinkAdapter*, std::unique_ptr<AudioTrackSinkAdapter>> m_mapAudioTrackAndSink;
```
kannan-xiao4, line 173: I think it can use `const` prefix.
```suggestion
        std::map<const AudioTrackSinkAdapter*, std::unique_ptr<AudioTrackSinkAdapter>> m_mapAudioTrackAndSink;
```

kannan-xiao4 (approved):

Issue #602, opened 2022-01-19
[BUG]: Latest develop branch version has significantly regressed in audio quality, send & receive audio has become choppy


rollersteaam:
### Package version

2.4.0-exp.5

### Environment

```markdown
* OS: Windows 10
* Unity version: Unity 2021.1.19f1
```

### Steps To Reproduce

1. Build your Unity WebRTC project to an Android and run it, checkout the latest commit in `develop` or checkout the commit `a8881cd4b06d740c999c885e6b5837b41d1b2d81` (the most recent commit I use in my test).
2. Connect to the same WebRTC call as your WebRTC browser (in my case, it is a different client that doesn't use this library, but implements the WebRTC protocol).
3. Perform an audio check. The Android project's receiving audio should be very choppy, and the browser side receiving audio should also be very choppy.
4. Optional: Check out commit `2b34a3a0066ae35a9e93d53a9ac899f214a4c44b` and repeat the whole test. Audio should have zero choppiness and be mostly smooth.

### Current Behavior

The Unity WebRTC project's send and receive audio is very choppy. It seems to take too long to send a frame of audio, as if it was trying to do it on the main thread and other Unity priorities like rendering were trying to happen first.

In the below audio tests, I do an audio check at 50% phone volume, 0% phone volume and 100% phone volume. Please note that phone volume is also creating large white noise from what seems like speaker feedback due to the onboard mic, which motivated me to file the noise suppression issue https://github.com/Unity-Technologies/com.unity.webrtc/issues/587. Using a bluetooth headset creates no white noise.

Audio test of the latest `develop` commit `a8881cd4b06d740c999c885e6b5837b41d1b2d81` (14th January 2022): https://www.loom.com/share/881e98f06d004927bd49c06b9d4cf656

Audio test of commit `2b34a3a0066ae35a9e93d53a9ac899f214a4c44b` (18th October 2021): https://www.loom.com/share/32f925920f274b6bb3e94fd2358a7dbe

### Expected Behavior

No send and receive audio choppiness. Ideally, also little to no white noise on 50% - 100% phone volume.

### Anything else?

Other clients in the WebRTC call hear the same audio quality (and artifacts) from the Unity WebRTC participant.

Perhaps relevant to note, this WebRTC setup is client-server where each participant receives all data from just one participant.

karasusan:
@rollersteaam 
Thanks for reporting. We will fix the issue soon.

karasusan:
[...]

PR #611, opened 2022-01-26
fix: Replace routine of periodic execution for audio buffer to fix audio latency

karasusan:
Related issue.
#525, #602

I investigated the issue of audio latency, and I found that there is a routine for processing audio buffer regularly, this doesn’t run at exact intervals.

I replaced this routine for periodic execution to [`RepeatingTaskHandle`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/rtc_base/task_utils/repeating_task.h).

### DIFF
--- a/Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
+++ b/Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
@@ -46,13 +46,19 @@ namespace webrtc
             int64_t elapsed_time_ms = -1;
             int64_t ntp_time_ms = -1;
 
-            const int kBytesPerSample = 2;
-            const int kChannels = 2;
-            const int kSamplingRate = 48000;
-            const int kSamplesPerFrame = kSamplingRate * kFrameLengthMs / 1000;
+            const size_t kBytesPerSample = 2;
+            const size_t kChannels = 2;
+            const size_t kSamplingRate = 48000;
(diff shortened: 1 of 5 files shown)


Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
```diff
@@ -46,13 +46,19 @@ namespace webrtc
...
-            const int kSamplingRate = 48000;
-            const int kSamplesPerFrame = kSamplingRate * kFrameLengthMs / 1000;
-
-            char data[kBytesPerSample * kChannels * kSamplesPerFrame];
+            const size_t kBytesPerSample = 2;
+            const size_t kChannels = 2;
+            const size_t kSamplingRate = 48000;
+            const size_t kSamplesPerFrame = kSamplingRate * kFrameLengthMs / 1000;
+            std::vector<int16_t> audio_data(kChannels * kSamplesPerFrame);
+            void* data = audio_data.data();
```
aet, line 54: hum, could initialize std::vector in the constructor once? :)

karasusan: Yes, this buffer is able to be on constructer. 
I will fix it. Thank you for your review!

karasusan: fixed it

kannan-xiao4 (approved):

PR #613, opened 2022-01-26
audio receiver with pull model rendering + audio conversion improvement

aet:
Hi!

I've been casually following the latest audio developments and happy to see all the progress and fixes.

I see there's also a new fix/audiothread branch in progress, but I wanted to throw out an alternative approach to discuss about. We could do away AudioClips and tracking, or circular buffer handling down in C# and using push model with callbacks before another audio thread actually fetches the data again from C#. :) I understand the need for callbacks with majority of the use cases but feel like audio shouldn't be one of them, streamline the pipeline.

I can't remember the internal implementation details of FMOD/OnAudioFilterRead in Unity but I would assume the benefits of using the pull model directly depends on the platform, like on a Mac/iOS CoreAudio is all pull model. My main testing platform is a Mac but testing on PC/Linux as well.

Most of this code was actually done over a year ago but I never got to contributing it. Instead of using OnData audio sinks, it was mixing all of receiving audio directly from audio_transport_ in DummyAudioDevice that suited my needs back then while waiting for the proper implementation.

I switched the audio float<>sint16 conversions to use webrtc built-in utillities. A while back there was some content where I was observing occasional audio tracks without the change.

Cheers,
-Antti

Issue #545, opened 2021-10-19
[BUG] Sending audio to two other peers is distorted on receiver's end


streunerlein:
**Describe the bug**
When sending ~~microphone input~~ _any audio_ (see my answer [below](https://github.com/Unity-Technologies/com.unity.webrtc/issues/545#issuecomment-947007716)) two other peers, the output on the third peer is highly distorted as if the sample rate was wrong. This is reproducible >50% of the time and is the case for local signaling as well as networked signaling with network delays. 

**To Reproduce**
1. Create a new Unity Project and import UnityWebRTC `2.4.0-exp4` with Examples
2. Open the `MultiplePeerConnections` scene in `Assets/Samples/WebRTC/2.4.0-exp4/Example/MultiplePeerConnections`
3. Replace the sample audio clip with microphone input as follows.

**Update:** You can also just use this clip, which is a microphone recording from inside Unity: https://dominiquesandoz.ch/share/test.wav

In *MultiplePeerConnectionsSample.cs*, replace the `Setup()` function:

```
// ...
        private void Setup()
        {
            Debug.Log("Set up source/receive streams");
            sourceStream = new MediaStream();

            var videoTrack = cam.CaptureStreamTrack(streamingSize.x, streamingSize.y, 0);
            sourceStream.AddTrack(videoTrack);
            sourceImage.texture = cam.targetTexture;

            sourceAudio.clip = audioclipStereoSample;
            sourceAudio.loop = true;
            sourceAudio.Play();
            var audioTrack = new AudioStreamTrack(sourceAudio);
            sourceStream.AddTrack(audioTrack);

            startButton.interactable = false;
            callButton.interactable = true;
        }
// ...
```

with

```
// ...
        private void Setup()
        {
            Debug.Log("Set up source/receive streams");
            sourceStream = new MediaStream();

            var videoTrack = cam.CaptureStreamTrack(streamingSize.x, streamingSize.y, 0);
            sourceStream.AddTrack(videoTrack);
            sourceImage.texture = cam.targetTexture;

            var localAudio = Microphone.Start("", true, 1, 48000);
    
            while (!(Microphone.GetPosition("") > 0)) {}

            sourceAudio.clip = localAudio;
            sourceAudio.loop = true;
            sourceAudio.Play();
            var audioTrack = new AudioStreamTrack(sourceAudio);
            sourceStream.AddTrack(audioTrack);

            startButton.interactable = false;
            callButton.interactable = true;
        }
[...]

PR #615, opened 2022-01-27
fix: Run all audio streaming process on the audio thread or the worker thread

karasusan:
Related issues #545, #525, #602
First, this pull request is affected from @aet suggestion [here](#613). So, he is a co-author of this pull request.
I appreciate his great contribution.

Unity provides the way of implementation to process something on audio thread. [`MonoBehaviour.OnAudioFilterRead(float[], int)`](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnAudioFilterRead.html) API is a callback which is called from the audio thread, and we can read/write the audio buffer before playing audio.

The basic idea is running tasks of audio streaming on the audio thread instead of the main thread. For this fix, playing audio smoothly even if the main thread has high load.

Additionally, we adopted the pull model for audio handling. The detail of design is described on the @aet 's pull request (#613).


aet:
I did some testing, sounds good to me! Thanks for cleaning up my old C style idioms from the code and simplifying it further, the original use case was a bit different. The audio render callback was actually calling transport_->NeedMorePlayData directly and you didn't necessarily get enough samples, the ringbuffer read/write all happened in the same thread. Writing the leftovers into ringbuffer etc.

I'm fine leaving OnAudioReceived alive, though I personally don't see where it could be useful :)

Some minor cleanups at https://github.com/aet/com.unity.webrtc/commit/09235adcf67984c56962188751af016e74cf8fc6 even if the callback only happens once, I'd prefer not dealing with the mutex before it's done

karasusan:
@aet 
Thanks for your review!

As you said, `OnAudioReceived` callback is not needed.
I fixed samples that depending the callback and I removed it.

I am experimenting that keeping shorter the ring buffer size.

aet:
One thing I didn't test was AudioStreamTrack(AudioSource source) code path. I see AudioSourceRead was removed but I'm not sure if AudioCustomFilter can replace it as-is? I was more interested getting rid of AudioClips elsewhere :) If it currently doesn't work as expected, maybe it can be changed to use a different variant of the filter and still rely on OnAudioFilterRead

Runtime/Scripts/AudioStreamTrack.cs
```diff
@@ -52,133 +52,70 @@ public class AudioStreamTrack : MediaStreamTrack
...
-                int lengthSamples = sampleRate;  // sample length for 1 second
-                return AudioClip.Create(
-                    clipName, lengthSamples, channels, sampleRate, false);
+                if (_filter != null)
+                    return;
+                _filter = GetOrAddComponent<AudioCustomFilter>(source.gameObject);
+                _filter.hideFlags = HideFlags.HideInInspector;
+                _filter.onAudioRead += SetData;
+                source.Play();
+                Debug.Log("AddFilter");
```
kannan-xiao4, line 114: I think no need this log in package script

Plugin~/WebRTCPlugin/WebRTCPlugin.h
```diff
@@ -25,12 +25,7 @@ namespace webrtc
     using DelegateSetSessionDescFailure =
         void(*)(PeerConnectionObject*, webrtc::RTCErrorType, const char*);
     using DelegateAudioReceive =
-        void(*)(AudioTrackSinkAdapter* sink,
-            const void* audio_data,
-            int size,
-            int sample_rate,
-            int number_of_channels,
-            int number_of_frames);
+        void(*)(AudioTrackSinkAdapter* sink);
```
kannan-xiao4, line 28: I think this definition is no needed.

Runtime/Scripts/AudioStreamTrack.cs
karasusan: fixed it

Plugin~/WebRTCPlugin/WebRTCPlugin.h
karasusan: fixed it

kannan-xiao4 (approved):

kannan-xiao4 (approved):

karasusan:
@aet AudioStreamTrack(AudioSource source) constructor is used by some samples, so I tested them but there is no problem.

Runtime/Scripts/AudioStreamTrack.cs
```diff
@@ -319,23 +186,23 @@ public AudioStreamTrack(AudioSource source)
                 throw new ArgumentNullException("AudioSource argument is null.");
             if (source.clip == null)
```
karasusan, line 187: @aet 
I may have misunderstood your comment.
Did you mean we can remove this line?

aet: No misunderstanding, glad to hear AudioCustomFilter works for both cases. I only realized later it was using a different SetData implementation!

I had the clip check removed before, but thought maybe it was there for a reason. I'm all forward removing it if no harm done

karasusan: I will make another PR for this fix.

aet:
Doesn't require updating the plugins, but noticed that Plugin~/WebRTCPlugin/AudioTrackSinkAdapter.cpp now includes common_audio/include/audio_util.h twice. easy to clean up at any point though

aet:
One more cleanup, unused public delegate void OnAudioReceived(AudioSource renderer) declared in Runtime/Scripts/AudioStreamTrack.cs

PR #619, opened 2022-01-28
Fix RTCStats.Type regression on Windows

aet:
The following code stopped triggering for me on PC once https://github.com/Unity-Technologies/com.unity.webrtc/pull/563 landed, while it still works on Mac/Linux where everything is built with Clang. I rarely use this functionality, so I didn't bother to debug this much further, but changing the type does help.

After changing the type, the plugin works the same if built with VS or Clang for PC
```
foreach (var rtpReceiver in pc.GetReceivers())
{
    RTCStatsReportAsyncOperation op = rtpReceiver.GetStats();
    yield return op;

    if (!op.IsError)
    {
        foreach (RTCStats stats in op.Value.Stats.Values)
        {
            RTCStatsType type = stats.Type;

            if (type == RTCStatsType.InboundRtp)
            {
                Debug.Log("RTCStatsType.InboundRtp")
            }
            if (type == RTCStatsType.Track)
            {
                Debug.Log("RTCStatsType.Track")
            }
        }
    }

    op.Value?.Dispose();
    op = null;
}
```

### DIFF
--- a/BuildScripts~/build_libwebrtc_android.sh
+++ b/BuildScripts~/build_libwebrtc_android.sh
@@ -29,7 +29,7 @@ mkdir -p "$ARTIFACTS_DIR/lib"
 
 for target_cpu in "arm64"
 do
-  mkdir "$ARTIFACTS_DIR/lib/${target_cpu}"
+  mkdir -p "$ARTIFACTS_DIR/lib/${target_cpu}"
 
   for is_debug in "true" "false"
   do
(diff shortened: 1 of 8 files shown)


karasusan (approved):

karasusan:
I reviewed changes and approved.
I do checkout your changes into our repository to test this change on our CI.

thank you for your contribution!

aet:
@karasusan there's probably a plugin update coming before the next release? just making sure things continue to work after native code change for Type

PR #627, opened 2022-02-10
linux vulkan loader fix

aet:
When I was updating Nvidia drivers on various Linux environments, I noticed that in some cases libvulkan.so and libcuda.so are simply missing and WebRTC crashes. But core unity renderer continues to get initialized as it's depending on libvulkan.so.1 instead on Linux and for some reason uses libvulkan.so explicitly for Android.

There's also a small error handling fix for current hardware encoder probe in case we have not detected a graphics device at all. Loading of vulkan driver fails, yet we shouldn't even get to this point as core unity has already moved forward using vulkan.

### DIFF
--- a/Plugin~/WebRTCPlugin/Dynlink/libcuda.so.init.c
+++ b/Plugin~/WebRTCPlugin/Dynlink/libcuda.so.init.c
@@ -25,7 +25,7 @@ extern "C" {
 
 #define CHECK(cond, fmt, ...) do { \
     if(!(cond)) { \
-      fprintf(stderr, "implib-gen: libcuda.so: " fmt "\n", ##__VA_ARGS__); \
+      fprintf(stderr, "implib-gen: libcuda.so.1: " fmt "\n", ##__VA_ARGS__); \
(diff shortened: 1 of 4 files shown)


karasusan:
Thanks for your contribution.
I will check it later.

aet:
@karasusan I removed the commits/comments about audio now that they are merged separately. But this landing this trivial PR would make loading the vulkan library consistent with Unity player code, so there wouldn't be an issue where Unity manages to load Vulkan support but WebRTC doesn't, leading to a crash

karasusan:
@aet 
Thanks for update the PR.
You mean there is another cause of crash?

aet:
Yes, on some Linux environments libvulkan.so does not exist so WebRTC fails from initializing. But Unity player loads against libvulkan.so.1 so you get into a situation where Unity tries to use WebRTC as Vulkan, but crashes later when loading the library fails

aet:
The PR is only about unifying against what filename the driver is loaded, matching Unity

kannan-xiao4 (approved):

PR #630, opened 2022-02-18
fix: RemoveSink method must be called before removing AudioTrackSink

karasusan:
This is a hot fix for crash bug about `AudioStreamTrack`.
This change is contained in #627 made by @aet.

### DIFF
--- a/Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
+++ b/Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
@@ -31,10 +31,15 @@ namespace webrtc
         if (!initialized_)
             return 0;
 
+        initialized_ = false;
+
+        taskQueue_->PostTask([this] {
+          task_.Stop();
+        });
+
         StopRecording();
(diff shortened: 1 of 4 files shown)


aet:
This reminds me of another bug I've ran into lately. When encoding multiple video tracks, the encoders are finalized at c# level but once it makes it down to UnityRenderEvent.cpp and VideoStreamRenderEventID::Finalize handling, it sometimes skips the release because ExistsRefPtr check for the track fails. Doesn't take too many attempts until you start seeing "encoder is null" warning messages.

I know the proper solution would be to get the new encoder branch finalized which doesn't do any of these mapping things :) I added a brute force workaround into UnityRenderEvent.cpp, but just in case something comes to your mind that would properly fix the issue

karasusan:
@aet
> I know the proper solution would be to get the new encoder branch finalized which doesn't do any of these mapping things :)
Yes, I can not wait for completing the work!

kannan-xiao4 (approved):

PR #638, opened 2022-03-01
fix slowly drifting audio desync + rely more on audio render thread

aet:
I've been told lately that audio desync starts to get noticeable after 30-40mins in our testcase that's a pretty heavily loaded app. Tried to look into what could be the regression from our previous M85 build from 13 months ago compared to what M92 webrtc has today.

It sounds like there's a chance for _convertedAudioData to build up slowly over time, so I took the chance to streamline the code similar to what was being done a year ago. I've tested this myself with over 4 hours of looping video to see if there's any desync, so far things seem to work but I'll get more feedback of the fix tomorrow.

### DIFF
--- a/Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
+++ b/Plugin~/WebRTCPlugin/DummyAudioDevice.cpp
@@ -63,30 +63,6 @@ namespace webrtc
                 kChannels, kSamplesPerFrame, data,
                 &elapsed_time_ms, &ntp_time_ms);
         }
-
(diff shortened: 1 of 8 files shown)


aet:
Heard from the testers it fixed the issue

karasusan:
@aet
Sorry for my late reply. We are grateful for testing.
We will check the fix in my environment. Thanks.

karasusan:
I created the new branch to rebase the PR to follow the latest update. #738

PR #644, opened 2022-03-04
fix PeerConnection/MediaStreamObserver leaks

aet:
I was wondering why UnityVideoTrackSource destructor is not called early enough and found these issues preventing it from being released until Context destructor clears them out.

### DIFF
--- a/Runtime/Scripts/Context.cs
+++ b/Runtime/Scripts/Context.cs
@@ -197,7 +197,7 @@ namespace Unity.WebRTC
 
         public void UnRegisterMediaStreamObserver(MediaStream stream)
         {
-            NativeMethods.ContextRegisterMediaStreamObserver(self, stream.GetSelfOrThrow());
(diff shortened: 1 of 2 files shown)


karasusan:
@aet
We are checking this fix. Thanks for your contribution!

kannan-xiao4 (approved):

karasusan:
memo: WRS-292

karasusan:
This pull request has the issue, so I fixed in another PR #712.

aet:
@karasusan interesting. I'm using this PR along with https://github.com/Unity-Technologies/com.unity.webrtc/pull/633, does it make any difference? Because I've never seen a crash

karasusan:
@aet Don't worry, your PR is no problem. Only it needed the patch for another crash bug.

karasusan:
We merged #712 so close this PR. Very thanks @aet!

PR #698, opened 2022-04-19
chore: Add TestRunnerOptions.json to set disable batchmode 

karasusan:
**TestRunnerOptions.json** file is for internal use. Internal CI system refer to this to read settings for testing.
Usually Unity Test Framework can not test audio features because it runs Unity as batchmode. **disableBatchMode** flag works to test features which are not able to be tested on batchmode.

### DIFF
--- a/TestRunnerOptions.json
+++ b/TestRunnerOptions.json
@@ -1,0 +1,3 @@
+{
+    "disableBatchMode": true
+}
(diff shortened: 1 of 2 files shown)


doctorpangloss:
Can you patch the engine instead to enable AudioManager under batch mode?

karasusan:
@doctorpangloss
I guess this option is not for runtime. 😓

kannan-xiao4 (approved):

PR #700, opened 2022-04-19
fix: Update UnityVulkanInitCallback.cpp

foijord:
Hook_vkGetInstanceProcAddr should apply vkGetInstanceProcAddr and return the result by default. As it is now, subsequent plugins that are loaded by Unity will get the Hook_vkGetInstanceProcAddr from this package, which returns nullptr for all but the specific functions checked for.

The change in this pull request is not tested in any way, and not even compiled.


karasusan (approved):

karasusan:
Thank you for the contribution!
We are checking and testing the pull request.

PR #721, opened 2022-05-10
fix: Fix error when using the video resolution which not supported by the H264 profile level

karasusan:
H.264 defines the maximum number of pixels and framerate depending on the profile level. The profile level used by most browsers is **3.1**, which does not support high-resolution frames such as Full HD. 

NvEnc will fail to initialize if an invalid profile level and resolution combination is specified. So as a workaround, it calculates a valid profile level for a given resolution and frame rate and uses it in NvEnc. Note that encoder initialization will fail if there is no valid profile level.

## TODO
Currently, the framerate used by NvEnc is a fixed value of **30** because the encoder framerate depends on the Unity frame rate. This will be supported in the future.

### DIFF
--- a/Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoderImpl.cpp
+++ b/Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoderImpl.cpp
@@ -27,10 +27,10 @@ namespace webrtc
 
     inline bool operator!=(const CUDA_ARRAY_DESCRIPTOR& lhs, const CUDA_ARRAY_DESCRIPTOR& rhs) { return !(lhs == rhs); }
 
-    inline absl::optional<webrtc::H264Level> H264RequiredLevel(const VideoCodec* codec)
+    inline absl::optional<webrtc::H264Level> H264RequiredLevel(const VideoCodec& codec)
     {
-        int pixelCount = codec->width * codec->height;
-        return H264SupportedLevel(pixelCount, static_cast<float>(codec->maxFramerate));
+        int pixelCount = codec.width * codec.height;
+        return H264SupportedLevel(pixelCount, static_cast<float>(codec.maxFramerate));
     }
 
     inline absl::optional<webrtc::H264Level> NvEncSupportedLevel(CUcontext context, const GUID& guid)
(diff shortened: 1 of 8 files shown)


kannan-xiao4 (approved):

Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoderImpl.cpp
```diff
@@ -408,22 +468,30 @@ namespace webrtc
...
 
-        // todo
-        fps = 30;
-
-        m_codec.maxFramerate = static_cast<uint32_t>(fps);
+        // workaround:
+        // Use required level if the profile level is lower than required level.
+        if (requiredLevel.value() > m_level)
+        {
+            m_level = requiredLevel.value();
```
kannan-xiao4, line 493: I think need to set m_level to `reconfigureParams` before reconfigure encoder.

Plugin~/WebRTCPlugin/Codec/NvCodec/NvEncoderImpl.cpp
```diff
@@ -408,22 +468,30 @@ namespace webrtc
...
         }
-        float fps = static_cast<float>(parameters.framerate_fps + 0.5);
-        int32_t pixelCount = m_codec.width * m_codec.height;
 
-        auto requireLevel = H264SupportedLevel(pixelCount, fps);
+        m_codec.maxFramerate = static_cast<uint32_t>(parameters.framerate_fps);
 
-        if (!requireLevel || m_level < static_cast<NV_ENC_LEVEL>(requireLevel.value()))
+        // todo(kazuki): Not supported framerate adjusting
+        m_codec.maxFramerate = 30;
```
kannan-xiao4, line 475: I think need to use `parameters` to determine the Bitrate and overwrite m_codec.maxBitrate.

```suggestion
        m_codec.maxFramerate = 30;
        m_bitrateAdjuster->SetTargetBitrateBps(parameters.bitrate.get_sum_bps());
        m_codec.maxBitrate = m_bitrateAdjuster->GetTargetBitrate();
```

kannan-xiao4 (approved):

EricUUU:
Hello, why is there no I frame when using H264 in the development branch? When switching large pictures, the remote picture of unity will get stuck, but there are no such problems when using VP8 format. @karasusan

PR #870, opened 2023-01-06
CI: Switch images used for running CI jobs

rasmus-unity:
### What
Changing images used in CI jobs

### Why
The `katana-to-yamato/*` images previously used will likely be deprecated at some point, so let's just have you now move to images which will be maintained going forward.

### Dependencies
None

### Note to reviewers
Looks like relevant CI jobs has been started automatically, hope I have covered everything

### DIFF
--- a/.yamato/nativeplugin.yml
+++ b/.yamato/nativeplugin.yml
@@ -4,7 +4,7 @@ build_platforms:
   - name: win
     type: Unity::VM
     flavor: b1.large
-    image: renderstreaming/win10:v0.3.13-1084239
+    image: renderstreaming/win10:v0.5.0-1259536
     build_command: BuildScripts~/build_plugin_win.cmd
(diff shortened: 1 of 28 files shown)


karasusan (approved):

rasmus-unity:
Hi @karasusan, given the failing jobs, I guess we shouldn't merge this PR. Tried to look at what the scripts are doing, but couldn't immediately figure out, so I let you handle it from here 🙂 
(Not super familiar with package building, but haven't seen other packages needing to copy files around like this in CI jobs, so perhaps it could be done in a different way?)

In any case, feel free to reach out if there is anything I can help with. The current `katana-to-yamato/*` images will likely live for a couple of more months at least, so this is not urgent.

rasmus-unity:
@karasusan, just to make sure that you are aware that eventually those `katana-to-yamato` images will be deprecated and stop working 🙂

PR #898, opened 2023-04-17
fix: Fix Encoded Transform samples

karasusan:
This pull requests fixes bugs of "metadata" sample witch is for demonstrating Encoded Transform. Previous implementation works VP8/VP9 codecs but not H264/AV1 codec. The issue of H264 codec has already fixed. However, The issue of AV1 codec is remained.

When using Encoded Transforrm API, replacing frame data has a limitation which we should not replace the first couple of bytes of the payload. For example, In [this script](https://github.com/webrtc/samples/blob/gh-pages/src/content/insertable-streams/endtoend-encryption/js/worker.js#L19) in the official webrtc sample repository contains comments about this limitation.

In addition, this pull request adds a "Encrypt" sample which shows a simple implementation of end-to-end encryption using Encoded Transform API.

https://user-images.githubusercontent.com/1132081/232367387-8e55e87c-28fb-428f-9297-ec55a57629f0.mp4

todo: Encoded Transforms doesn't work with AV1 codec.

### DIFF
--- a/Samples~/Encryption/EncryptionSample.cs
+++ b/Samples~/Encryption/EncryptionSample.cs
@@ -1,0 +1,571 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using Unity.Collections;
+using Unity.WebRTC;
+using Unity.WebRTC.Samples;
+using UnityEngine.UI;
+using Button = UnityEngine.UI.Button;
+
+class EncryptionSample : MonoBehaviour
+{
+#pragma warning disable 0649
+    [SerializeField] private Button startButton;
+    [SerializeField] private Button callButton;
+    [SerializeField] private Button restartButton;
+    [SerializeField] private Button hangUpButton;
+
+    [SerializeField] private Camera cam;
+    [SerializeField] private RawImage sourceImage;
+    [SerializeField] private RawImage receiveImage;
+    [SerializeField] private Transform rotateObject;
+    [SerializeField] private InputField encryptKeyInput;
+    [SerializeField] private InputField decryptKeyInput;
+#pragma warning restore 0649
+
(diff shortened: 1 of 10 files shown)


Samples~/Encryption/EncryptionSample.cs
```diff
@@ -0,0 +1,571 @@
...
+        receiver.Transform = new RTCRtpScriptTransform(TrackKind.Video, e => OnReceiverTransform(receiver.Transform, e));
+    }
+
+    void OnSenderTransform(RTCRtpTransform transform, RTCTransformEvent e)
+    {
+        lock (encryptKeyInputLock)
+        {
+            if(encryptKeyArray.IsCreated)
+            {
+                if(e.Frame is not RTCEncodedVideoFrame frame)
```
kannan-xiao4, line 188:
```suggestion
                if (!(e.Frame is RTCEncodedVideoFrame frame))
```
`is not` syntax does not support on Unity 2020.3.

Samples~/Encryption/EncryptionSample.cs
```diff
@@ -0,0 +1,571 @@
...
+    }
+
+    void OnReceiverTransform(RTCRtpTransform transform, RTCTransformEvent e)
+    {
+
+        lock (decryptKeyInputLock)
+        {
+            if (decryptKeyArray.IsCreated)
+            {
+                if (e.Frame is not RTCEncodedVideoFrame frame)
```
kannan-xiao4, line 224:
```suggestion
                if (!(e.Frame is RTCEncodedVideoFrame frame))
```
same as https://github.com/Unity-Technologies/com.unity.webrtc/pull/898/files#diff-4b5680394927d27eb23ee24f50df9a36851d684113eb0e1c120dc0075170c46bR188

kannan-xiao4 (approved):

PR #902, opened 2023-04-21
Fix cuda resource leak

aet:
Mostly just code cleanups for CUDA handling, but should fix resource leaks for D3D12/Vulkan that use mipmappedArray. M92 era branch used to handle calling cuMipmappedArrayDestroy

Can't be sure if this alone fixes all the leaks I've seen, as I'm still using custom frame pooling, scheduling and vulkan changes on my branch, but have to start from somewhere.

### DIFF
--- a/Plugin~/WebRTCPlugin/GpuMemoryBuffer.cpp
+++ b/Plugin~/WebRTCPlugin/GpuMemoryBuffer.cpp
@@ -38,7 +38,13 @@ namespace webrtc
 #endif
     }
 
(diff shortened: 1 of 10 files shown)


BrianHarrisonUnity:
Looks good to me.
When I get a chance I will test this to confirm it resolves the cuda resource leaks I was seeing, and approve.

Thanks!

kannan-xiao4 (approved):
LGTM!

karasusan (approved):

PR #916, opened 2023-05-18
Chore: Add .mm files to clang-format

aet:
Format Objective-C++ files with clang-format as well, no functional changes


karasusan (approved):
It looks good to me

PR #930, opened 2023-06-29
fix: Support scale resolution for h264 codec

karasusan:
- Add CUDA kernel for scaling down texture resolution.
- Add **enable_language(CUDA)** in CMakeLists.txt to build CUDA kernel.
- Upgrade **CUDA Toolkit 11.6**, CUDA supports VS2022 since 11.6. 
- Change **msvc** compiler for windows plugin because **clang-cl** is not worked with CUDA compiler.

### DIFF
--- a/Plugin~/NvCodec/Utils/ResizeSurf.cu
+++ b/Plugin~/NvCodec/Utils/ResizeSurf.cu
@@ -124,7 +124,7 @@ CUresult ResizeSurf(CUarray srcArray, CUarray dstArray)
         srcWidth / dstWidth,
         srcHeight / dstHeight);
 
-    result = cuSurfObjectDestroy(dstSurface);
+    result = cuSurfObjectDestroy(srcSurface);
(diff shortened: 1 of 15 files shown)


Plugin~/NvCodec/Utils/ResizeSurf.cu
```diff
@@ -0,0 +1,131 @@
...
+        srcSurface,
+        srcWidth,
+        srcHeight,
+        dstSurface,
+        dstWidth,
+        dstHeight,
+        srcWidth / dstWidth,
+        srcHeight / dstHeight);
+
+    result = cuSurfObjectDestroy(dstSurface);
```
kannan-xiao4, line 127:
```suggestion
    result = cuSurfObjectDestroy(srcSurface);
```
maybe `srcSurface` destory? (not dstSurface)

kannan-xiao4 (approved):

Issue #803, opened 2022-09-01
[REQUEST]: Improve input latency for streaming from Unity camera to end users


doctorpangloss:
### Is your feature request related to a problem?

This is an approach to improve the latency with the current architecture of the plugin.

The goal is to get the frame encoded within 7ms of finishing rendering (https://parsec.app/blog/nvidia-nvenc-outperforms-amd-vce-on-h-264-encoding-latency-in-parsec-co-op-sessions-713b9e1e048a), which is pretty close to NvFBC / Moonlight.

Right now, due to the architecture of the plugin, the time between rendering finishing and encoding the frame is about
 - 7ms for NVIDIA to encode
 - 4-10ms to copy resources
 - 0-20ms due to frame rate control
 - 0-24ms due to WaitForEndOfFrame
 totalling about 11-61ms of input latency when networking RTT is 0.

### Describe the solution you'd like

 - the plugin event call should be implemented as a custom pass in HDRP, a render feature in URP and the current method (end of frame) for the obsolete pipeline
 - Input System should queue events directly from the data channel's thread instead of using `WebRTC.Sync`. this will require a fix for Input System's memory leak with its input event buffers (https://forum.unity.com/threads/bug-memory-leaks-crash-when-queueing-events-from-threads-other-than-the-main-thread.1329837/)
 - video frame rate control should be turned off
 - the render texture passed to webrtc must be retained on c# side until webrtc reports it is done encoding it.
 - gpu memory buffers should be created with a reference to the `ITexture2D*`. then, they should create the appropriate "texture view" for the hardware/software encoder on demand inside the encoder queue thread. this means `handle()` does the `map` step in the current implementation, and `ToI420` gets the cpu texture.
 - in order to avoid the expensive `map` on windows, there must be DX11 and DX12 NVEncoderImpl support, instead of using CUDA encoder for all platforms.

doctorpangloss:
some other architectural ideas:
[...]

PR #950, opened 2023-07-12
performance: RTCRtpSender.SyncApplicationFramerate

karasusan:
This implements Frame synchronization mode.
https://github.com/Unity-Technologies/com.unity.webrtc/issues/803

### DIFF
--- a/Plugin~/WebRTCPlugin/UnityVideoTrackSource.cpp
+++ b/Plugin~/WebRTCPlugin/UnityVideoTrackSource.cpp
@@ -25,7 +25,7 @@ namespace webrtc
         taskQueue_ = std::make_unique<rtc::TaskQueue>(
             taskQueueFactory->CreateTaskQueue("VideoFrameScheduler", TaskQueueFactory::Priority::NORMAL));
         scheduler_ = std::make_unique<VideoFrameScheduler>(taskQueue_->Get());
-        scheduler_->Start(std::bind(&UnityVideoTrackSource::CaptureNextFrame, this));
+        scheduler_->Start(std::bind(&UnityVideoTrackSource::OnUpdateVideoFrame, this));
     }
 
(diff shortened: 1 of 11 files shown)


Plugin~/WebRTCPlugin/UnityVideoTrackSource.cpp
```diff
@@ -20,11 +20,12 @@ namespace webrtc
         : AdaptedVideoTrackSource(/*required_alignment=*/1)
         , is_screencast_(is_screencast)
         , frame_(nullptr)
+        , syncApplicationFramerate_(false)
     {
         taskQueue_ = std::make_unique<rtc::TaskQueue>(
             taskQueueFactory->CreateTaskQueue("VideoFrameScheduler", TaskQueueFactory::Priority::NORMAL));
         scheduler_ = std::make_unique<VideoFrameScheduler>(taskQueue_->Get());
-        scheduler_->Start(std::bind(&UnityVideoTrackSource::CaptureNextFrame, this));
+        scheduler_->Start(std::bind(&UnityVideoTrackSource::OnUpdateVideoFrame, this));
```
vl4dimir, line 28: Shouldn't this line be wrapped with `if (!syncApplicationFramerate_)`?

Or alternatively, `Start` unconditionally but immediately `Pause` if `syncApplicationFramerate_` is `true`? Since I see you're kind of relying on scheduler to be active, in the `SetSyncApplicationFramerate` function below.

karasusan: @vl4dimir
OK, I fixed the constructor to pause the scheduler at first and make `syncApplicationFramerate_` true by default.

vl4dimir:
@karasusan Thanks for fixing this! Is there a plan for when this will be packaged so we can try it out?

PR #955, opened 2023-07-19
feat: simulcast H264 nvcodec

karasusan:


aet:
Glad to see the scale resolution change landed. I had been sitting on a similar functionality to contribute forward, but it was still depending on NPP and didn't have time to remove the extra dependency.

For NV simulcast, there was an experiment repeating the simulcast bits directly into NvEncoderImpl but ultimately it didn't seem to bring any added value vs adding the support through EncoderSimulcastProxy with less maintenance and code changes, did you consider using the proxy earlier? I can share my changes if needed, just in progress rebasing my code against latest version regardless.

karasusan:
Thanks for advice!

We previously considered using NPP, but decided to implement it in the CUDA kernel because we did not want to add more dependencies.

I am trying to implement simulcast, but my current implementation is not working yet. There are performance issues. If you have already tried this, I would appreciate it if you could let me know.

I haven't ever seen EncoderSimulcastProxy, it looks already deleted on the chromium repository.
https://source.chromium.org/chromium/_/webrtc/src.git/+/f6eae959bf9b26476a8ed12ba5d2a631903a1f79

aet:
:) It's part of M112 at media/engine/encoder_simulcast_proxy.h and everything still seems to be compiling while I'm rebasing against the second latest revision from main, without the scale resolution PR. I'll try get it cleaned up against scale resolution PR this week so I can send something over to test

aet:
@karasusan got the basics rebased against the latest main and seems to still work at least on Linux player, I haven't had time to do bigger editor runtime testing while migrating from M107 to M112 yet

PR #957, opened 2023-07-21
refactor: Refactor CUDA handle

karasusan:
Refactored to consolidate CUDA handle generation for all Graphics APIs in one place.

### DIFF
--- a/Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
+++ b/Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
@@ -392,11 +392,6 @@ namespace webrtc
         if (!IsCudaSupport())
             return nullptr;
 
-        GMB_CUDA_CALL_NULLPTR(cuCtxPushCurrent(GetCUcontext()));
-
-        std::unique_ptr<GpuMemoryBufferCudaHandle> handle = std::make_unique<GpuMemoryBufferCudaHandle>();
-        handle->context = GetCUcontext();
-
         VulkanTexture2D* vulkanTexture = static_cast<VulkanTexture2D*>(texture);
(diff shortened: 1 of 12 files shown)


Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
```diff
@@ -405,37 +405,9 @@ namespace webrtc
             RTC_LOG(LS_ERROR) << "cannot get export handle";
             return nullptr;
         }
-
```
aet, line 408: I think the handle is still created here but not used anymore? I was creating the handle early and using it against the cu* calls directly, so that in case there are any errors and part of the initialized stuff has to be withdraw, the destructor of the GpuMemoryBufferCudaHandle handles that.

karasusan: Sorry, I failed to rebase to the main branch. I am fixing it.

aet: :) Sorry for nitpicking, there's two cases left where GpuMemoryBufferCudaHandle gets created late vs original code

karasusan: Could you tell me where there's two cases?

karasusan: Sorry but I might misunderstood your advice. Can you see me the patch?

karasusan: Thanks, cherry-picked it.

kannan-xiao4 (approved):

PR #959, opened 2023-07-26
fix: fractional scaling ratio support

aet:
Finished testing of using #930 vs my previous implementation using NPP and noticed a small regression.

- It seems scaleResolutionDownBy = 1.5 doesn't quite work, trivial casting issue
- ~~The image quality is a bit blurry, so I added a lanczos alternative and image quality is now comparable to what I've been using with NPP~~ EDIT: Kernel worked in local testing, but don't have time to debug an issue when running on cloud. For now, I'll just fallback to my old NPP scaling code in internal use

### DIFF
--- a/Plugin~/NvCodec/Utils/ResizeSurf.cu
+++ b/Plugin~/NvCodec/Utils/ResizeSurf.cu
@@ -2,7 +2,7 @@
 
 #include <cuda_runtime.h>
 
-__global__ void ResizeSurfNearestNeiborKernel(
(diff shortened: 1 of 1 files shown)


aet:
@karasusan still testing this, potentially found some issues that could get fixed by porting code from NPP branch, just without use of NPP itself

aet:
@karasusan ready for review, for now I just left the fractional scaling ratio fix in to play it safe

karasusan (approved):

PR #962, opened 2023-08-16
Fix: Cuda thread safety with D3D11

aet:
I was doing some testing on PC with Unity editor, though I'm mostly using Vulkan.

I ran into occasional crash issue with D3D11 where the device gets lost and no matter what I did, couldn't repro it with other graphics apis. Someone else had similar issue at https://forums.developer.nvidia.com/t/d3d11-device-context-in-a-separate-thread-gets-corrupted-when-cuda-graphics-resource-mapping-is-used/232326 and luckily applying similar locking fixed the issue for me

I don't recall seeing this issue before, but seems the chances of this happening started after the recent fence change


karasusan (approved):

karasusan:
@aet
Thanks you for checking the crash and fixing it. 
It works on my environment.

PR #967, opened 2023-08-30
fix: Vulkan CPU texture read performance, improve Unity pipeline state integration

aet:
With the use of IUnityGraphicsVulkan::CommandRecordingState out of the way, here's another part I've been using in my branch for few years now. The plugin architecture has changed few times over the years, so I never had time to try pitching it in. 😄 

- VulkanTexture2D::GetNativeTexturePtrV returns the real native texture handle like other targets

- DoImageLayoutTransition tweaks to avoid touching layout and to remove most if not all Vulkan API validation issues

- Query if VK_MEMORY_PROPERTY_HOST_CACHED_BIT available, improves CPU encoder performance. It was tested against 1080p60 at the time, but most of my testing goes against nvenc

- Prep work for supporting staging textures (cpu writes instead of cpu reads preference), though it's missing VkBuffer related additions. This will be used for improving video rendering performance

- Add GetTexturePtr helper. I need a way to figure out from a native texture handle if it's maintained by Unity or the plugin. For now it's Vulkan only, but I'm using it with other targets when adding native texture uploads/staging texture support.


aet:
@karasusan this might indirectly help with the android thing

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanUtility.cpp
```diff
@@ -364,7 +375,9 @@ namespace webrtc
             {
                 barrier.srcAccessMask = VK_ACCESS_HOST_WRITE_BIT | VK_ACCESS_TRANSFER_WRITE_BIT;
             }
-            barrier.dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
+
+            // Vulkan API validation error?
+            // barrier.dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
```
BrianHarrisonUnity, line 380: I noticed these validation errors as well, but I believe the mask is correct.
Really strange!

aet: Agreed! At this point it's been a long time I was fine-tuning the details so my memory is a bit hazy, but I was trying to find common ground that'll work with the way Unity graphics device handles the layout/state flushes (across multiple versions) and something that was also working with RTX/GTX drivers

aet: @BrianHarrisonUnity @karasusan I refreshed my memory on this part a bit and tested against recent enough Unity LTS (2021.3.24ff1) in case something in the Vulkan implementation has changed since the original implementation time. It looks like whatever made me disable the dstAccessMask at the time, it's now safe to revert back.

It's unfortunate that Unity plugin interface doesn't fill `UnityVulkanImage.layout` for us when calling `AccessTexture`, so you'll have to make some assumptions. But this should only be relevant when uploading a texture from the plugin towards Unity. I've kept this this extra fiddling out of the PR for simplicity sake, it's more relevant to video rendering performance with staging texture support.

When running Unity player build on Linux with -force-vulkan-layers, I'm only getting two validation errors during initialization but they're not related to WebRTC.
```
VUID-vkCmdClearAttachments-rect-02682(ERROR / SPEC): msgNum: -1380104337 - Validation Error: [ VUID-vkCmdClearAttachments-rect-02682 ] Object 0: handle = 0x7f97d00b98f0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xadbd476f | CmdClearAttachments(): pRects[0].rect.extent.width is zero. The Vulkan spec states: The rect member of each element of pRects must have an extent.width greater than 0 (https://vulkan.lunarg.com/doc/view/1.3.250.1/linux/1.3-extensions/vkspec.html#VUID-vkCmdClearAttachments-rect-02682)
VUID-vkCmdClearAttachments-rect-02683(ERROR / SPEC): msgNum: -566636251 - Validation Error: [ VUID-vkCmdClearAttachments-rect-02683 ] Object 0: handle = 0x7f97d00b98f0, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0xde39d125 | CmdClearAttachments(): pRects[0].rect.extent.height is zero. The Vulkan spec states: The rect member of each element of pRects must have an extent.height greater than 0 (https://vulkan.lunarg.com/doc/view/1.3.250.1/linux/1.3-extensions/vkspec.html#VUID-vkCmdClearAttachments-rect-02683)
```

karasusan:
@aet
I haven't tried **VK_MEMORY_PROPERTY_HOST_CACHED_BIT** bit. Does it improve the performance?
I'm profiling it.

aet:
@karasusan it's been a while since I've used Vulkan with software encoder path, but back then I was testing against software vp8/vp9 and h264 encoders and was wondering why I couldn't reach 60fps 1080p easily even with h264 on a Linux machine with plenty of cores.

It turned out reading the texture with the Vulkan implementation was really slow at VulkanGraphicsDevice::ConvertRGBToI420 and adding VK_MEMORY_PROPERTY_HOST_CACHED_BIT when it's available helped. But it's hard to make apples to apples comparison with the main branch, I've been using a custom frame scheduling and queuing for so long now.. they were tuned for performance at the time with sacrificing latency or memory use a bit. Little by little trying to get changes contributed forward that make sense.

karasusan:
@aet
OK, if I can be of any help to test, please let me know.

aet:
@karasusan hands-on quick test in android could be useful :) There's a lot of generic things outside of Vulkan that could be altered to avoid dropping frames, but it doesn't hurt to have shorter CPU read times. I'm trying to refresh my old debug notes from 2+ years ago, but it does seem like the CPU read for me took 22ms at 1080p and 10ms for smaller resolutions.. after fix, less than 1ms each.

This was around M85 and there was that custom RGB to YUV conversion code that was slowing down the process a bit, but got converted to use libyuv directly later on. But I think I was measuring the read and conversion separately.

karasusan:
@aet
Right, first I will measure the effect of the VK_MEMORY_PROPERTY_HOST_CACHED_BIT flag in my environment.

aet:
@karasusan I got over my laziness to measure my own code after a long break 😄 

On LinuxStandalone, for testing sake I'll disable h264 nvenc entirely and fallback to h264 software encoder that I've enabled for my libwebrtc build.

With m_hasHostCachedMemory = false the average time it takes to read 1080p texture with vkMapMemory/vkUnmapMemory/webrtc::I420Buffer::Create and libyuv::ARGBToI420 calls is around 40ms.

With m_hasHostCachedMemory it's 0.75ms on the average for me.

Tested with Intel i7-9750H / Geforce RTX 2060.

karasusan:
@aet
I checked this fix on Android device, and it can determine this makes dramatically performance improvement. In special, the result of software encoding like VP8 is noticeable. 

At the same time, however, I found the video image disturbance when setting vSync enable on Android. This issue might be occurred since #966. I am keeping the investigation to fix the issue.

aet:
@karasusan I wasn't sure what was the main motivation getting CommandRecordingState into use in PR 966 after PR 888 closed? Not that I don't want it to happen, just wasn't sure if there was more reasons. :)

I am able to reproduce the same WaitSync regression from the original PR 888 when vsync is enabled. There's not much hooks to work with from the plugin towards Unity rendering state, so synchronizing against frame counts is always kind of hacky compared to custom CommandBuffer and VkFence. On Linux, running 60fps app with vsync gives me about 20fps stream and WaitSync was either done at 18ms or timing out at 30ms.

Since Unity Vulkan graphics device updates the safe frame counter once a frame, after presenting, we'll need to add some C# side PlayerLoop extras that get called after Unity's built-in PostLateUpdate that syncs the graphics state on plugin side.

I have a small prototype of this, that fixes the framerate back to 60fps with vsync and LinuxStandalone. It's an add-on to PR 966 that doesn't conflict with this PR. I'll try to get it out for testing tomorrow, it's likely also improving the reason why WaitSync was timing out and causing crash on Android in PR 966 that had to be worked around.

I'm not sure what you mean by image disturbance, but if it's not related to this PR, I'm ok moving forward to merge. I'll add the vsync changesets to another PR, unless you prefer adding them here

aet:
@karasusan please do me a favor and instead of reverting all of #966, let's just restore the part that really matters, texture specific cmdbuf/fence synchronization. I added a new changeset into this PR to allow using either cmdbuf/fence or CRS for synchonization, so it's easier to debug and compare this going forward once there's more time to look into the vsync issue.

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.h
```diff
@@ -80,6 +84,11 @@ namespace webrtc
         VkCommandPool m_commandPool;
         VkCommandBuffer m_commandBuffer;
         VkFence m_fence;
+#else
+        static void AccessQueueCallback(int eventID, void* data);
```
BrianHarrisonUnity, line 88: I think with Kazuki's revert, we also removed AccessQueueCallback.
This should be able to be removed right?

Plugin~/WebRTCPlugin/UnityRenderEvent.cpp
```diff
@@ -104,8 +103,7 @@ static void UNITY_INTERFACE_API OnGraphicsDeviceEvent(UnityGfxDeviceEventType ev
             UnityVulkanPluginEventConfig batchUpdateEventConfig;
             batchUpdateEventConfig.graphicsQueueAccess = kUnityVulkanGraphicsQueueAccess_DontCare;
             batchUpdateEventConfig.renderPassPrecondition = kUnityVulkanRenderPass_EnsureOutside;
-            batchUpdateEventConfig.flags = kUnityVulkanEventConfigFlag_EnsurePreviousFrameSubmission |
-                kUnityVulkanEventConfigFlag_ModifiesCommandBuffersState;
+            batchUpdateEventConfig.flags = kUnityVulkanEventConfigFlag_ModifiesCommandBuffersState;
```
BrianHarrisonUnity, line 107: We will need an #ifdef around the event config since it changes depending on which version we use I think.

I think it should be like the following when using VULKAN_USE_CRS:
```cpp
            UnityVulkanPluginEventConfig batchUpdateEventConfig;
            batchUpdateEventConfig.graphicsQueueAccess = kUnityVulkanGraphicsQueueAccess_DontCare;
            batchUpdateEventConfig.renderPassPrecondition = kUnityVulkanRenderPass_EnsureOutside;
            batchUpdateEventConfig.flags = kUnityVulkanEventConfigFlag_EnsurePreviousFrameSubmission |
                kUnityVulkanEventConfigFlag_ModifiesCommandBuffersState;
```

I think it should be like the following when using !VULKAN_USE_CRS:
```cpp
            UnityVulkanPluginEventConfig batchUpdateEventConfig;
            batchUpdateEventConfig.graphicsQueueAccess = kUnityVulkanGraphicsQueueAccess_Allow;
            batchUpdateEventConfig.renderPassPrecondition = kUnityVulkanRenderPass_EnsureOutside;
            batchUpdateEventConfig.flags = kUnityVulkanEventConfigFlag_EnsurePreviousFrameSubmission;
```

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.h
aet: Yup, it was removed in the original PR 966 but now that we try to keep both options (CRS vs separate cmdbuf/fence) open for additional testing and debugging, I added it back. I wasn't sure if there's any firm decisions either way for using AccessQueueCallback, but I can cherry-pick changesets here if needed or further tweaks could be another PR from Kazuki?

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
```diff
@@ -176,215 +227,246 @@ namespace webrtc
...
+        }
+
+        if (m_unityVulkan != nullptr)
+        {
+            m_unityVulkan->AccessQueue(AccessQueueCallback, 0, texture, false);
+        }
+        else
+        {
+            AccessQueueCallback(0, texture);
+        }
```
BrianHarrisonUnity, line 249: We will need to swap this to Kazuki's updated version.
We didn't just revert the entire change, but also removed the AccessQueueCallback piece.

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.cpp
```diff
@@ -19,6 +19,10 @@ namespace unity
 {
 namespace webrtc
 {
+#if !VULKAN_USE_CRS
+    static VulkanGraphicsDevice* s_GraphicsDevice = nullptr;
+#endif
```
BrianHarrisonUnity, line 24: I think this can go as well actually.
This was for the AccessQueueCallback, but with removing that we can get rid of this as well I believe.

Plugin~/WebRTCPlugin/GraphicsDevice/Vulkan/VulkanGraphicsDevice.h
BrianHarrisonUnity: I think we need to cherry pick / merge in the further changes from Kazuki during the revert.
The updated branch, "origin/fix/revert-vulkan", is how I think we can fix the original issue that lead to #966. 

I am pretty sure AccessQueue & the two part event was causing the issue that lead to the #966 fix, and the updated fix branch should work, but also needs testing to ensure that we don't hit issues trying to copy the resource Unity owns / controls with having exclusive queue access configured on the event.

Ideally, I think this branch should be a swap between the two if possible.

BrianHarrisonUnity: This commit, "52cf51c7ad4bff35f0648fd0a4823772adcbc28b", is the one we have to port.
I think there is some additional clean-up we could do with AccessQueue removed, but can ignore that for now.

aet: Thanks! Just noticed the same changeset and branch, so crafted the QueueSubmit changes and AccessQueueCallback removal from there. I've never been a fan of kUnityVulkanEventConfigFlag_EnsurePreviousFrameSubmission with the CRS path so I left that one out, but once there's time to do more comparison and debugging, we can get back to it.

BrianHarrisonUnity:
Thanks for the quick updates!

I'll give it a test when I get a chance, but I think this looks good.

karasusan:
@aet 
I tested it on Android and Linux. It looks good to me.
Thanks you for contribution.

Issue #964, opened 2023-08-22
[BUG]: The sound is captured back to the microphone


MaximKurbanov:
### Package version

3.0.0-pre.6

### Environment

```markdown
* OS: Android
* Unity version:2021.3

Hi!
I have big problem with my voice chat App. When 2 people voice each other by mobile phone. The voice sound of the speaker is recorded back to the microphone, and causing echo. After many times the sound is echoed and resonant with each other, causing loud noises. I've been looking for solutions like AEC(Acoustic Echo Cancellation), Echo Suppression, but it's not work for Unity.
Have you ever had this problem or has any solutions, please let me know. Thanks so much!!!
```

### Steps To Reproduce

Run app on 2 mobile devices

karasusan:
@MaximKurbanov 

#651 
Is this the same issue?

Joshua-Douglas:
Hello, 

Thank you so much for your tireless efforts with this package @karasusan! My team has been using this package extensively and we've found that it works very well. I'm wondering if this bug might be a good opportunity for our team to contribute back to this repository.

I've noticed that the current implementation hard-codes the `AudioOptions` within the `WebRTCPlugin` `Context` object.

```c++
rtc::scoped_refptr<AudioSourceInterface> Context::CreateAudioSource()
    {
        // avoid optimization specially for voice
        cricket::AudioOptions audioOptions; // echo_cancellation will default to nullptr
        audioOptions.auto_gain_control = false;
        audioOptions.noise_suppression = false;
        audioOptions.highpass_filter = false;
        return UnityAudioTrackSource::Create(audioOptions);
    }
```

Do you think `noise_suppression`, `echo_cancellation`, etc would work as expected if these options were set to `true`? If so, then we might be able to close out this issue by exposing the AudioOptions and allowing end-users to specify `echo_cancellation` if they need it (this would also close #587 I believe).

If you think this would be a good approach, and if you are open to contributions, then I would be happy to submit a fix. I'd be happy to outline the changes in a design document before submitting a PR, so you could approve the changes before I begin implementing.

karasusan:
@Joshua-Douglas
Hi, always we welcome to your contribution. Making these audio options to public is a first step to solve this issue. I'm not sure these opitions can fix it, therefore we need to test these options.

If you make a PR, we will be glad to review and test.

Joshua-Douglas:
Hello @karasusan, that is great to hear - thanks! 

[...]

PR #978, opened 2023-09-21
add audiooption

karasusan:


flamacore:
So hey, I've followed the issues that are mentioned in this PR and I also would like to give a try to the audio options.

Sadly I have problems with getting this specific pr to build as I'm not very well versed with CMake and the like. It throws an error about not finding libraries and shuts down. 

Anyhow, is there any progress on making this a reality? Any sort of easier way I can add this and experiment with it in my own project yet?

karasusan:
@flamacore
Hi, if you can help us to test the audio option, I'll be glad to take you a package. 

Here you are.
https://drive.google.com/file/d/1Qf4ZLTqNxT-hoLfgJj3JZIpFu3hEPLNI/view?usp=drive_link

flamacore:
@karasusan would be delighted to. Thanks a lot for sharing the package. I'll give it a check soon and will inform if it works or not.

MaximKurbanov:
I tested and issue #964 remains. 
I tried this with android-android and windows-android:
```cs
  #...
    {
  #...
        var track = new AudioStreamTrack(sendAudioSource, GetAudioOptions());
        mediaStream.AddTrack(track);
    }

    private AudioOptions GetAudioOptions()
    {
        return new AudioOptions()
        {
            echoCancellation = true,
            noiseSuppression = true
        };
    }
```

flamacore:
Can confirm the echo issue persists. Not sure if it's related to us enabling/disabling the options or if the internal echo cancellation works as intended.

karasusan:
@MaximKurbanov 
@flamacore

Thank you for checking. I'd like to clarify which this issue is a specific by mobile platforms or not. 
For example, windows and windows, or windows or macOS.

It looks there are different implementations for mobile and desktop platforms. 
https://stackoverflow.com/questions/62479789/webrtc-android-echo-cancellation

MaximKurbanov:
@karasusan 
I tested Windows-Mac, AEC does not work.

karasusan:
@MaximKurbanov
I understood the cause of this issue.
Sorry but we need more time to fix it.

I can't share you many information about that, but at least I can say that the "audio processing" in native code is not processed.

The "audio processing" in webrtc is implemented here.
https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/audio_processing/audio_processing_impl.cc

pixlhero:
@karasusan 
I tested the package from Oct 6th on Mac-Editor <-> iPad.
It also didn't cancel the echo.

This is a show-stopper for us. So we really, really need this fix in the next release (pre-8). 🙏

Also building the plugin doesn't work for us. So if 02724602838a48b95bac2e1389d03c10c39b3c7e would fix anything (we're targeting iOS) it would be great if you could provide us with a .tgz build.

pixlhero:
> @karasusan I tested the package from Oct 6th on Mac-Editor <-> iPad. It also didn't cancel the echo.
> 
> This is a show-stopper for us. So we really, really need this fix in the next release (pre-8). 🙏
> 
> Also building the plugin doesn't work for us. So if [0272460](https://github.com/Unity-Technologies/com.unity.webrtc/commit/02724602838a48b95bac2e1389d03c10c39b3c7e) would fix anything (we're targeting iOS) it would be great if you could provide us with a .tgz build.
> 
> がんばってよ!

We found a workaround on iOS involving `AVAudioSession`. So for our project this issue is resolved. @karasusan where should we post the workaround?

Issue #972, opened 2023-09-07
[BUG]: Bug or feature?


MaximKurbanov:
### Package version

3.0.0-pre.6

### Environment

```markdown
* OS: All
* Unity version: 2021+
```

### Steps To Reproduce

By default, when sending and receiving video, the picture flips vertically.

Sender:
//Flip vertically
`Graphics.Blit(sourceTexture_, destTexture_, s_verticalScale, s_verticalOffset);`
Receiver:
and receiver flip back
`m_renderer = new UnityVideoRenderer(this, needFlip: true);`

### **It is not possible for the plugin user to change the auto flip of the receiver**
```
/// <summary>
/// Video Receiver
/// </summary>
/// <param name="ptr"></param>
internal VideoStreamTrack(IntPtr ptr)
    : base(CreateVideoTrack(ptr))
{
    if (!s_tracks.TryAdd(self, new WeakReference<VideoStreamTrack>(this)))
        throw new InvalidOperationException();

    m_dataptr = Marshal.AllocHGlobal(Marshal.SizeOf(typeof(VideoStreamTrackData)));
    Marshal.StructureToPtr(m_data, m_dataptr, false);

    m_renderer = new UnityVideoRenderer(this, true); <---
}
```

karasusan:
@MaximKurbanov 


VideoStreamTrack provides a constructor with a `needFlip` argument. Can you try this?

```
public VideoStreamTrack(Texture texture, bool needFlip = true)
```

MaximKurbanov:
@karasusan 
This constructor for sender!

MaximKurbanov:
Fixed #971

PR #971, opened 2023-09-07
feat: Added support for horizontal flip sourceTexture

MaximKurbanov:
When I added support for "VUZIX M400 SMART GLASSES" to the project, the picture from the glasses was diagonally flipped. So I added the horizontal and diagonal flip.

```
videoStreamTrack = new VideoStreamTrack(webCamTexture, false, true); //Flip horizontaly
sourceRawImage.uvRect = new Rect(1, 1, -1, -1); //Flip diagonally for display
```

PR #981, opened 2023-09-25
fix: Fix flip vertical video image for received video

karasusan:


aet:
the default behavior for VideoStreamTrack changed too. Now i'm supposed to pass CopyTextureHelper.VerticalFlipCopy, is it on purpose?

karasusan:
The previous PR changes the flag false. It looks passing true is right.
https://github.com/Unity-Technologies/com.unity.webrtc/pull/971/files#diff-f4a6f4c4bccd5c3f082e2ee65fc6785d7c1af35098d1d68374a1ae2955ba7f25R182

MaximKurbanov:
> @aet The previous PR changes the flag false. It looks passing true is right. https://github.com/Unity-Technologies/com.unity.webrtc/pull/971/files#diff-f4a6f4c4bccd5c3f082e2ee65fc6785d7c1af35098d1d68374a1ae2955ba7f25R182

I also changed the Sender, now the Sender sends the image without flip. And upon receipt, a flip is not required. If we put flip, we will get a vertically flipped image. 
If you receive a normal image from a camera(webcamera) without flip, then you also do not need to flip it when sending. Else if you received a fliped image from a camera, then you need to flip it to normal and send it. So the receiver doesn't need flip.
#972

aet:
@MaximKurbanov Hum, I don't use any of the webcam stuff but just receiving external video tracks and rendering to textures and sending them forward (includes the received video track), in both scenarios the default behavior changed? I think the intent of the new CopyTexture parameter is good, to give flexibility to rare cases like the smart glasses you were using.

The default should be something that works for majority of the users, being backwards compatible is a plus. I'm ok explicitly adding the flip flags on my tracks, but seems like extra step for majority of the users.

aet:
Sorry, I was talking about `public VideoStreamTrack(Texture texture, bool needFlip = true)` changing into `VideoStreamTrack(Texture texture, CopyTexture copyTexture = null)` that's different from `NeedReceivedVideoFlipVertically`

MaximKurbanov:
> The default should be something that works for majority of the users, being backwards compatible is a plus. I'm ok explicitly adding the flip flags on my tracks, but seems like extra step for majority of the users.
I agree with you.

MaximKurbanov:
My recommendation for optimization is to set NeedReceivedVideoFlipVertically to false. 
And use "copyTexture: Graphics.Blit" in the VideoStreamTrack constructor.

karasusan:
@MaximKurbanov
I agree with you about suggestion of the performance improvement. However, it is needed to support browser and unity at the same time.

Issue #896, opened 2023-04-13
[BUG]: The encoder sometimes chooses the wrong minimum level and causes a parameter exception


doctorpangloss:
### Package version

3.0.0-pre.4

### Environment

```markdown
* OS: Windows 2022
* Unity version: 2021 LTS
```

### Steps To Reproduce

1. Request a 1688x614 frame, frame rate 60, max bit rate 24000000, target bps 4925000
2. Observe the level computing code incorrectly computes Level 4 for this request.
3. Observe it should be at least 4.2.
4. Observe an exception.
5. `m_level = std::max(m_level, NV_ENC_LEVEL_H264_42);` resolves the issue.

Windows 2022 LTSC + DirectX 11 + NVEnc streaming to Windows Chrome: Works
Windows 2022 LTSC + DirectX 11 + NVEnc streaming to Mobile Safari, Portrait: Works
Windows 2022 LTSC + DirectX 11 + libvpx streaming to Mobile Safari, Portrait: Works
Windows 2022 LTSC + DirectX 11 + libvpx streaming to Mobile Safari, Landscape: Works
Windows 2022 LTSC + DirectX 11 + NVEnc streaming to Mobile Safari, Landscape: Does not work (see parameters to repro)

### Current Behavior

An invalid parameter exception thrown by `m_encoder->CreateEncoder(&m_initializeParams);`

### Expected Behavior

Select Profile 4.2 instead of 4.

### Anything else?

The workaround is to always set the level to at least 4.2.

kannan-xiao4:
@doctorpangloss 
Thank you for reporting. We investigate this issue.
memo: WRS-467

Can you tell me in which environment it occurs and in which environment there is no problem?

doctorpangloss:
Windows 2022 LTSC + DirectX 11 + NVEnc streaming to Chrome: Works
Windows 2022 LTSC + DirectX 11 + NVEnc streaming to Mobile Safari, Portrait: Works
Windows 2022 LTSC + DirectX 11 + libvpx streaming to Mobile Safari, Portrait: Works
Windows 2022 LTSC + DirectX 11 + libvpx streaming to Mobile Safari, Landscape: Works
Windows 2022 LTSC + DirectX 11 + NVEnc streaming to Mobile Safari, Landscape: Does not work (see parameters to repro)

When using the particular parameters I wrote there, it should reproduce without needing to go through this whole rigamarole.

kannan-xiao4:
[...]

PR #1049, opened 2024-09-12
fix: h264 level selection algorithm

RebootVanChild:
There seems to be a slight issue in the h264 level selection algorithm. Widths and heights of frames and h264 macroblocks should be taken into account when doing the calculation, not only the pixel counts.
This issue can cause encoder invalid param error on certain resolution ranges.

For example, a 1922x1080 30fps video stream:

The correct algorithm should be
```
macroblocksPerFrame = Ceil( pixelsFrameWidth / pixelsMacroblockSide ) * Ceil( pixelsFrameHeight / pixelsMacroblockSide )
= Ceil( 1922 / 16 ) * Ceil( 1080 / 16 )
= 8228
```
Which resides in level 4.2.

However, according to current algorithm

```
macroblocksPerFrame = Ceil( pixelsPerFrame / pixelsPerMacroblock )
= Ceil( 1922 * 1080 / (16 * 16) )
= 8109
```
Thus, the algorithm would select level 4, and cause encoder invalid param error.

### DIFF
--- a/Plugin~/WebRTCPlugin/Codec/H264ProfileLevelId.cpp
+++ b/Plugin~/WebRTCPlugin/Codec/H264ProfileLevelId.cpp
@@ -38,20 +38,23 @@ namespace webrtc
         { 2073600, 36864, 240000, H264Level::kLevel5_2 },
     };
 
-    static const int kPixelsPerMacroblock = 16 * 16;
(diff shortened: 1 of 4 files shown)


doctorpangloss:
is this related?

https://github.com/Unity-Technologies/com.unity.webrtc/issues/896

Really the minimum level should always be h264 42

PR #1062, opened 2024-12-05
doc: Improve documentation of RTCIceServer

miyabi:


Runtime/Scripts/WebRTC.cs
```diff
@@ -483,32 +483,42 @@ public enum RTCIceCredentialType
     }
 
     /// <summary>
-    ///
+    ///     Represents a configuration for an ICE server used within WebRTC connections.
     /// </summary>
-    /// <seealso cref="RTCConfiguration"/>
+    /// <remarks>
+    ///     `RTCIceServer` struct represents a configuration for an ICE server used within WebRTC connections,
```
chrstphrsxtn, line 489:
```suggestion
    ///     Represents a configuration for an ICE server used within WebRTC connections,
```
As per the [guidance](https://docs.unity.com/internal/content-ops/en-us/documentation/content-model/api-content-model/cs-engine-struct#summary), you don't need to repeat the struct's name here (it will be clear from the context which API element is being described) , but this is a minor point of course.

chrstphrsxtn (approved):
Approving with one minor style suggestion.

PR #1101, opened 2025-09-10
Fix: Segfault in encoder on platforms that use I420 frame buffer type.

Kleptine:
This PR fixes a segfault on MacOS, and likely other platforms that utilize the I420 fallback path. The encoder thread will segfault if more than 60ms of work is bottlenecked on the GPU. 

In the I420 pipeline, when requesting an I420 frame, a GPU -> CPU copy is queued up, and then waited on in WaitSync(). There is a default timeout set to 60ms, in which case WaitSync simply returns false and a nullptr. This helps avoid stalling the encoder thread, however this case isn't handled properly by the UnityVideoEncoder, causing it to submit a webrtc::VideoFrame with a nullptr as a data buffer. When the encoding thread receives this, it attempts to read from that null frame buffer and segfaults.

**Fix:** This eagerly checks the result of GetI420() in order to make sure it is valid before sending it along to the encoder. We only do this on kI420 typed frame buffers, so that paths with GPU based textures don't accidentally enqueue an additional copy.

Happy to make any modifications to the approach -- let me know if you have any thoughts.

**Reproduceability**: 100%
**Tested**: MacOS Silicon

### DIFF
--- a/Plugin~/WebRTCPlugin/UnityVideoEncoderFactory.cpp
+++ b/Plugin~/WebRTCPlugin/UnityVideoEncoderFactory.cpp
@@ -75,6 +75,19 @@ namespace webrtc
         int32_t Release() override { return encoder_->Release(); }
         int32_t Encode(const VideoFrame& frame, const std::vector<VideoFrameType>* frame_types) override
         {
+            // Eagerly check for a valid I420 buffer to handle GPU timeouts.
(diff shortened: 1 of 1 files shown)


Kleptine:
Hi, any update on either of these fixes? Would love to be able to update my package dependencies.

